Issue #66, opened 2020-02-18
Add simple animations


Chrriis:
FlatLaf _looks_ modern, but it could _feel_ more modern if it had subtle animations.

Potential areas:
- Tab selection indication (underline).
- Buttons (e.g. ripple effect on click).
- Scroll bar (current trend is to paint a small width scrollbar and to paint it fully on mouse hover. An animation makes it full width).
- Showing/hiding tick marks on check boxes with a smooth transition.
- Tree collapse/expand arrows that rotate.

smileatom:
Not to crash this party but personally I feel that java based L&Fs have a ton of issues that have never been addressed.  Differentiators like (what makes your app better than mine), would be awesome, but Java GUIs have alot of issues.  Things like close buttons on tabs, draggable tabs yeah ok but things still dont look act right by default.

Chrriis:
@smileatom I am not sure I understand. You seem to complain about the drawbacks of Swing in general, which a look and feel is not going to solve.
A look and feel is about the _look_ and the _feel_. FlatLaf looks good but does not feel slick, especially compared to modern applications which users now expect (generally developed with web technologies, CSS transitions, etc.). It is not possible for us to alter the feel unless we patch FlatLaf. Improvements like animations would not just make my app look better, but all apps using FlatLaf. 

As for close buttons on tabs, draggable tabs, these are nice to have but are doable externally to the look and feel. Of course, it could help if the look and feel offered assistance for this, at the cost of being tied to that specific look and feel (e.g. Tri-state checkbox).

smileatom:
I actually have no problems with swing other than looking good and the many attendant border/focus/font/functionality issues from a UX perspective.  Thats why I continue to use it.  I think your probably doing more advanced swing development than I have been (I abandoned swing many years ago because its not commercially viable), but this project could change that.  Im hoping for the best ;)

PR #67, opened 2020-02-18
Animation

Chrriis:
This is a naive implementation of animation for tab selection. It is a way to trigger discussion and get advice with regards to issue #66.

Even if it were considered a step in the right direction, it is not complete: there should be a setting to make the animation feature on/off (in some complex layouts, animations may not be desirable as it repaints stuff that may take time to render).


DevCharly:
Thanks for the PR. I'll test it next week because I'm out-of-office until Wednesday.

PR #68, opened 2020-02-21
FlatButtonUI getPreferredSize nullPointer fix

basix86:
FlatButtonUI.getPreferredSize throw exception when button.getComponentCount()>0.
In that case we should return null as same way of BasicButtonUI.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java
@@ -398,7 +398,7 @@ public class FlatButtonUI
 		if( isHelpButton( c ) )
 			return new Dimension( helpButtonIcon.getIconWidth(), helpButtonIcon.getIconHeight() );
 
-		Dimension prefSize = super.getPreferredSize(c);
+		Dimension prefSize = super.getPreferredSize( c );
 		if ( prefSize == null )
 			return null;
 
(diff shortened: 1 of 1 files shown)


smileatom:
public Dimension getPreferredSize(JComponent c)
Description copied from class: ComponentUI
Returns the specified component's preferred size appropriate for the look and feel. If null is returned, the preferred size will be calculated by the component's layout manager instead (this is the preferred approach for any component with a specific layout manager installed). The default implementation of this method returns null.
Overrides:
getPreferredSize in class ComponentUI
Parameters:
c - the component whose preferred size is being queried; this argument is often ignored, but might be used if the UI object is stateless and shared by multiple components

DevCharly:
Thanks for the fix. Good catch 👍 

I've checked the other usages of `super.getPreferredSize(JComponent c)` and added a check to `FlatRadioButtonUI`. The other usages never return `null`.

DevCharly:
@basix86 and @smileatom
Just noticed that you're not show in the list of contributors here:
https://github.com/JFormDesigner/FlatLaf/graphs/contributors

The reason is probably that the email address you use for github.com is not equal to the email address in your git commits. Would you be so kind and add your git commit email address as additional email address to github.com. You can do this when you log in to github.com, go to "Settings > Email", fill the field "Add email address" and click the "Add" button. This does not change your primary github.com email address. Then you should be listed on the contributors page. Thanks.

basix86:
@DevCharly Thanks! I have fixed the email and now I am visibile in the list of contributors

Issue #70, opened 2020-02-26
Improve disabled button rendering


basix86:
The look and feel render the disabled button with a grayscale.

FlatLaf inherit getDisabledIcon method from "BasicLaF".
It has hight performance (https://stackoverflow.com/a/9131751) but I think that we can choose a better graphic solution.

"BasicLaF" internally use a GrayFilter with a 50% value filter.
I have tested many values to have a lighter solution. The best values are near 70 and 80.
The problem with these high value filter is that we lose a color contrast.

I have seen that "MetalLaF" override the basic method and I think that the graphical result is better.

See here an example:
 - on top the actual Flat and "Basic" solution
- in the bottom the "Metal" solution

![disabledButton](https://user-images.githubusercontent.com/17149962/75368606-4acbee80-58c2-11ea-9ac1-d85dc1d037f1.jpg)

If you accept my idea, I can prepare a merge request with a solution inspired from "MetalLaF"(MetalUtils.getOceanDisabledButtonIcon).

With this solution we can also add to the UIManager the key "Button.disabledGrayRange" = [180,215]

DevCharly:
Yes, this looks better 👍 
A PR would be great.

Will this work for the dark themes too?
With different gray range?

BTW the FlatLaf .properties files do not support arrays. So it is probably easier to use two integer UI values. E.g. `Button.disabledGrayRangeMin` and `Button.disabledGrayRangeMax`.
With [FlatUIUtils.getUIInt( String key, int defaultValue )](https://github.com/JFormDesigner/FlatLaf/blob/master/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatUIUtils.java#L109) its easy to get integers and pass defaults.

basix86:
I have created the PR here:
https://github.com/JFormDesigner/FlatLaf/pull/72

I have tested with various themes, ad it work well.
With dark theme we suggest to change the value of grays from [180-215] to a lower values, for example [1-100].

This solution is very pretty with gray themes but with colored theme should be possible to replace a gray scale with a single color scale

DevCharly:
@basix86 I've merged PR #72 into new branch 'disabled-icons'.

For testing with some real world icons I've created a little app `com.formdev.flatlaf.testing.FlatDisabledIconsTest`

The disabled icons look good. Maybe a little bit to light.
But they are hard to see on selected disabled toggle buttons.

[...]

Issue #71, opened 2020-02-27
Share Java Codestyle


basix86:
Hi,
@DevCharly can you share your java code-style file to align all developers commits?

DevCharly:
Sure, what IDE do you use?

basix86:
IntelliJ IDEA

DevCharly:
@basix86 I've just added code style for IntelliJ IDEA to `.editorconfig`. This should be automatically used by IDEA in the FlatLaf project.

Requires IDEA EditorConfig plugin, which is bundled with IDEA (since some versions) and enabled by default.

DevCharly:
Also added formatter profile for Eclipse.

PR #72, opened 2020-02-27
Improve disabled button rendering #70

basix86:


flatlaf-core/src/main/java/com/formdev/flatlaf/FlatLaf.java
```diff
@@ -36,20 +36,13 @@
...
-import javax.swing.JLabel;
-import javax.swing.JRootPane;
-import javax.swing.JTabbedPane;
-import javax.swing.LookAndFeel;
-import javax.swing.PopupFactory;
-import javax.swing.SwingUtilities;
-import javax.swing.UIDefaults;
-import javax.swing.UIManager;
-import javax.swing.UnsupportedLookAndFeelException;
+import javax.swing.*;
```
smileatom: This is probably not desired.  Explicit imports is preferred.

flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatUIUtils.java
```diff
@@ -16,28 +16,19 @@
...
-import java.awt.Container;
-import java.awt.Dimension;
-import java.awt.Font;
-import java.awt.Graphics;
-import java.awt.Graphics2D;
-import java.awt.Insets;
-import java.awt.Rectangle;
-import java.awt.RenderingHints;
-import java.awt.Shape;
+import java.awt.*;
```
smileatom: Again likely undesirable

smileatom: Again likely undesirable

flatlaf-core/src/main/java/com/formdev/flatlaf/FlatLaf.java
basix86: ok, I wait your codestyle
https://github.com/JFormDesigner/FlatLaf/issues/71

DevCharly:
Thanks for the PR. I'll test it next week because I'm out-of-office until Wednesday.

flatlaf-core/src/main/java/com/formdev/flatlaf/FlatLaf.java
smileatom: Although .* imports seems like a code style issue.. Its not especially in the case of root level UI code like L&F.  This should be reverted.

PR #88, opened 2020-04-25
Fix relative resource location

matt-pan:
Using flatlaf intellijthemes as a dependency does not work for all themes from the subfolder 'material-theme-ui-lite' because of the usage of ".." to  to load the resource.

### DIFF
--- a/flatlaf-intellij-themes/src/main/java/com/formdev/flatlaf/intellijthemes/materialthemeuilite/Utils.java
+++ b/flatlaf-intellij-themes/src/main/java/com/formdev/flatlaf/intellijthemes/materialthemeuilite/Utils.java
@@ -32,7 +32,7 @@ class Utils
 
 	static IntelliJTheme loadTheme( String name ) {
 		try {
-			return new IntelliJTheme( Utils.class.getResourceAsStream( "../themes/material-theme-ui-lite/" + name ) );
+			return new IntelliJTheme( Utils.class.getResourceAsStream( "themes/material-theme-ui-lite/" + name ) );
 		} catch( ParseException | IOException ex ) {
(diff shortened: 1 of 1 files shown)


DevCharly:
Hmm, can't reproduce the problem.

Can you give me some details about your environment?

- used Java version
- is Java module system?
- some kind of plug-in system used? or custom classloader?

DevCharly:
Forget my previous comment. I can reproduce it now.

matt-pan:
... just as I prepared an example :)
Thing is - the resource load always searches from the package root, not the class root - thus the "step to parent with .." from the Utils class is not needed

DevCharly:
Hmm, actually the ".." should be correct because the materials `Utils` class is in package
`com.formdev.flatlaf.intellijthemes.materialthemeuilite`. Having a resource
`../themes/material-theme-ui-lite/Arc Dark Contrast.theme.json` gives
`com/formdev/flatlaf/intellijthemes/materialthemeuilite/../themes/material-theme-ui-lite/Arc Dark Contrast.theme.json` which should result in
`com/formdev/flatlaf/intellijthemes/themes/material-theme-ui-lite/Arc Dark Contrast.theme.json`

It worked in the dev environment, but it seems that `Class.getResourceAsStream()` works slightly different when invoked from a JAR...

Anyway, also had problems without ".." so I changed it to absolute paths in 129bc9b3ae79136c7a0f00d594f3112e98f594f4

Could you test the latest `0.33-SNAPSHOT` snapshot? 
https://github.com/JFormDesigner/FlatLaf#snapshots

matt-pan:
Works from within the IDE and from the finished build as JAR!

Issue #82, opened 2020-04-12
TitleBar color


koldou:
Depending on these selected ColorSchme, set the JFrame TitleBar colour, the default background color.

DevCharly:
Yes, that would be great, but I don't think that it is possible to change TitleBar color in pure Java because it comes from the operating system.

A solution would be to re-implement the whole window title in Java (issue #47).
Then we have full control, but probably loose OS specific behavior and look.

DevCharly:
this is now released in [FlatLaf 0.37](https://github.com/JFormDesigner/FlatLaf/releases/tag/0.37)

Issue #47, opened 2020-01-15
Look-and-feel-decorated JFrames do not render properly


Chrriis:
(split from issue #39)

Modern applications that use flat widgets may want their frame decorations also to be flat.

Here is a test case of the current situation:
```Java
	public static void main(String[] args) throws Exception {
		UIManager.setLookAndFeel(new FlatLightLaf());
		JFrame frame = new JFrame("JFrame");
		frame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
		frame.setUndecorated(true);
		frame.getRootPane().setWindowDecorationStyle(JRootPane.FRAME);
		JDesktopPane desktopPane = new JDesktopPane();
		JInternalFrame internalFrame1 = new JInternalFrame("JInternalFrame1", true, true, true, true);
		internalFrame1.setLocation(50, 30);
		internalFrame1.setSize(300, 200);
		desktopPane.add(internalFrame1);
		JInternalFrame internalFrame2 = new JInternalFrame("JInternalFrame2", true, true, true, true);
		internalFrame2.setLocation(60, 40);
		internalFrame2.setSize(300, 200);
		desktopPane.add(internalFrame2);
		frame.getContentPane().add(desktopPane);
		frame.setSize(400, 300);
		frame.setVisible(true);
		internalFrame1.setVisible(true);
		internalFrame2.setVisible(true);
		internalFrame2.setIcon(true);
	}
```

... and here is the result:
![FrameDecorations](https://user-images.githubusercontent.com/1174202/72375936-4fbb5100-370d-11ea-872a-d1eab44b5cdf.png)

@DevCharly said:
Window decoration style looks similar to internal frames, but it seems that the implementation has nothing to do with internal frames. It is implemented in javax.swing.plaf.metal.MetalRootPaneUIand javax.swing.plaf.metal.MetalTitlePane.

Poivin:
I agree, it would be very interesting to have a "flatLAF" decoration on Jframes and Jdialogs in order to give the possibility to completely open the title bars implementation through UIs for customisation and through Overriding for specific features. @DevCharly Is it in FlatLAF roadMap ? Thanks for all your work!

Chrriis:
I thought having a L&F decorated frame was secondary, but in fact the default title bar totally ruins the dark mode experience. Here is a screenshot of a sample app:
[...]

PR #108, opened 2020-06-02
WIP: Custom window decorations

DevCharly:
This PR adds custom window decorations to FlatLaf, which supports dark window title panes and embedding menu bar into title pane. (issues #47 and #82)

This is work in progress.

Current state:
- title pane implemented in Windows 10 style with:
  - window icon and title
  - minimize, maximize, restore and close buttons
- move window works
- resizing window works
- only Windows 10 style
- only supported on Windows 10
- **native Windows 10 borders** when running in [JetBrains Runtime 11](https://confluence.jetbrains.com/display/JBR/JetBrains+Runtime) or later (the JRE that IntelliJ IDEA uses)
  - native drop shadows
  - native window resize and move
  - native window snapping

To use custom window decorations in your application, enable them with following code (before creating any frames or dialogs).
~~~java
JFrame.setDefaultLookAndFeelDecorated( true );
JDialog.setDefaultLookAndFeelDecorated( true );
~~~

If running on Windows 10, custom window decorations are used and the **menu bar is automatically moved into the title pane**. On other platforms or older Windows versions, the custom window decorations are not used.

Java 8 with menu bar embedded:
image

Java 8 with menu bar not embedded:
image

Dark:
image

Running in JetBrains Runtime 11:
image

image


k33ptoo:
Commendable work, is this already available to test?

DevCharly:
@k33ptoo this is now released in [FlatLaf 0.37](https://github.com/JFormDesigner/FlatLaf/releases/tag/0.37)

k33ptoo:
I have tested and it's pretty amazing thank you and good job.

grimlock81:
Great work, this looks cool on Windows 10! The only drawback is with a dark theme the border around the JFrame is a whitish colour and this really stands out especially when the frame is maximised. I don't have a JetBrains Runtime 11 to compare, but I assume this is what you meant when you wrote 
> native Windows 10 borders when running in JetBrains Runtime 11 or later (the JRE that IntelliJ IDEA uses)
Is there no way to change the frame's border colour?

Also, the title on the JFrame being rendered after the menu bar, makes it look weird as it is uncentered. I know this is following the fact that without the menu bar the title would have been rendered with a LEFT horizontal alignment by Windows natively, but since this is now being rendered by FlatTitlePane, is it possible to add a setting to make the titleLabel rendered with a CENTER horizontal alignment?

tsjung75:
When I turn on this feature, I encounter a problem.
I am using a heavyweight component (OCX, wrapped in Java) and add it, but sometimes it disappears suddenly.
Actually, I add this component dynamically (remove from one parent panel and add to other parent panel).
If I turn off this feature, it is OK.

DevCharly:
@tsjung75 have no idea...

Does the problem occur with embedded menu bar?
Please run your app with VM arg `-Dflatlaf.menuBarEmbedded=false` to test whether it works without embedded menu bar.

DevCharly:
@grimlock81 window border color is currently taken from Window system (vie desktop property "win.frame.activeBorderColor") (and there is no way to change it). Thought that it is a good idea to use color from Window system, which the user can change in Windows Settings dialog, but this was a bad decision. It does not work and it actually makes no sense to get border color from Windows but use own color for title pane. I'll change this for next release...

Centering title is good idea. I'll add add a setting. Thanks.

DevCharly:
@grimlock81 that the border is shown when the window is maximized is a bug. The border should be hidden if window maximized.

tsjung75:
@DevCharly I don't use the menu bar in my application. I will stick to not using this flag for a while. Thanks again~

DevCharly:
@grimlock81 window border color is now fixed in master branch. Also the window title is now centered and maximized windows do not have a border.

DevCharly:
@grimlock81 forgot to mention that the window border color now comes from UI default `RootPane.activeBorderColor` for active windows and from `RootPane.inactiveBorderColor` for inactive windows.

Chrriis:
@tsjung75 I think native components have problems painting when using `JFrame.setDefaultLookAndFeelDecorated(true)`.

tsjung75:
> @tsjung75 I think native components have problems painting when using `JFrame.setDefaultLookAndFeelDecorated(true)`.

You're right. I'll not use this option for application that has native components

grimlock81:
@DevCharly Just tried it on 0.38, the window border color fix is perfect and title centering looks good, just what I needed. Thanks.

grimlock81:
I've found an issue when using custom window decorations on Windows 10 when programmatically maximizing the JFrame by calling `frame.setExtendedState(Frame.MAXIMIZED_BOTH);`

The resulting JFrame's dimensions fits the whole screen including covering the Windows task bar. This is inconsistent with how Windows applications are displayed in Maximized - they don't cover the task bar. The behaviour of subsequent calls to `frame.setExtendedState(Frame.MAXIMIZED_BOTH);` depends on whether the Maximize button (next to the Close button) has been clicked at least once in the same session or not:

- If the Maximize button has been clicked, then the call correctly sets the dimensions of the JFrame to fit the screen but doesn't cover the task bar
- If the Maximise button has never been clicked, then the call incorrectly sets the dimensions to cover the task bar.

This is my test code
```
import com.formdev.flatlaf.FlatIntelliJLaf;

import java.awt.EventQueue;
import java.awt.Frame;
import java.awt.event.ActionEvent;
import javax.swing.AbstractAction;

import javax.swing.JFrame;
import javax.swing.JMenu;
import javax.swing.JMenuBar;
import javax.swing.JScrollPane;
import javax.swing.JTree;
import javax.swing.UIManager;

/**
 *
 * @author grimlock81
 */
public class TestExtendedState {
    public static void main(String[] args) {
        EventQueue.invokeLater(() -> {
            startGUI();
        });
    }

    private static void startGUI() {
        JFrame.setDefaultLookAndFeelDecorated(true);
        try {
            UIManager.setLookAndFeel(new FlatIntelliJLaf());
        }
        catch (Exception ex) {
            ex.printStackTrace();
        }

        JFrame frame = new JFrame();

        JMenuBar menuBar = new JMenuBar();
        JMenu fileMenu = new JMenu("File");
        fileMenu.add("Open"); // fake
        fileMenu.add("Exit"); // fake
        JMenu editMenu = new JMenu("Set");
        editMenu.add(new AbstractAction("Extended State to Normal") {
            @Override
            public void actionPerformed(ActionEvent evt) {
                frame.setExtendedState(Frame.NORMAL);
            }
        });
        editMenu.add(new AbstractAction("Extended State to Maximized Both") {
            @Override
            public void actionPerformed(ActionEvent evt) {
                frame.setExtendedState(Frame.MAXIMIZED_BOTH);
            }
        });
        menuBar.add(fileMenu);
        menuBar.add(editMenu);

        frame.setJMenuBar(menuBar);
        frame.setTitle("Testing Extended State");
        frame.add(new JScrollPane(new JTree()));
        frame.pack();
        frame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
        frame.setVisible(true);
        frame.setExtendedState(Frame.MAXIMIZED_BOTH);
    }
}

```

grimlock81:
After looking through FlatTitlePane.java I can understand why the behaviour is correct after pressing the Maximize button, because `frame.setmaximizedBounds(Rectangle)` is called and sets the right dimensions accounting for the task bar. Does that mean that Windows is incorrectly setting the maximized bounds at the start?

DevCharly:
@grimlock81 could you please open a new issue for the maximized problem?
(this PR is closed and GitHub does not support reopening it)

iamabs2001:
My Class extends to JFrame then how i can set  
```java
JFrame.setDefaultLookAndFeelDecorated(true);
```
i tried in main method and constructor as well but not getting custom decorated window :( help me.

Issue #200, opened 2020-10-30
TextField: select all text on mouse click


basix86:
Hi, I have tested the property on FlatLaf Demo
`TextComponent.selectAllOnFocusPolicy=always`
but the _select all_ works only when I enter in text filed pressing tab button on keyboard.

I have seen the code in the class FlatCaret

```
	@Override
	public void focusGained( FocusEvent e ) {
		if( !wasTemporaryLost && !isMousePressed )
			selectAllOnFocusGained();
		wasTemporaryLost = false;
		wasFocused = true;

		super.focusGained( e );
	}
```

To have the selection on mouse click I have tried to remove the `isMousePressed  ` check.

I don't know if is safe to remove it. Can be added another boolean property to enable this case?

DevCharly:
The reason/idea for the `isMousePressed` check is that when the user clicks somewhere at the text, I assume that he wants to put the caret at that position. That's also how Windows (win32) text boxes seem to work:
- gain focus with tab key selects all text
- gain focus via mouse click sets caret at clicked position and clears selection

Fixed with PR #202. Thx.

PR #202, opened 2020-11-04
TextField: allow select all on mouse click

basix86:
See issue here:
https://github.com/JFormDesigner/FlatLaf/issues/200


DevCharly:
LGTM. Thanks

PR #208, opened 2020-11-11
Range slider ui

basix86:
Jide RangeSliderUi support


basix86:
The FlatRangeSliderUI extends FlatSliderUI and override paintTrack method (the fields now are protected).
Now the color track is always between the two values.

All other methods are copy/paste from BasicRangeSliderUI. It is not the best solution, but Jide have used the same approach for every look and feel (MetalRangeSliderUI, WindowsRangeSliderUi, ecc..).

I have also added the test launcher, It is my first experience with JFormDeisigner and I don't know if I used it in the correct way

PR #209, opened 2020-11-11
Add RangeSlider support

basix86:
The FlatRangeSliderUI extends FlatSliderUI and override paintTrack method (the fields now are protected).
Now the color track is always between the two values.

All other methods are copy/paste from BasicRangeSliderUI. It is not the best solution, but Jide have used the same approach for every look and feel (MetalRangeSliderUI, WindowsRangeSliderUi, ecc..).

I have also added the test launcher, It is my first experience with JFormDeisigner and I don't know if I used it in the correct way

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatSliderUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatSliderUI.java
@@ -61,14 +61,14 @@ import com.formdev.flatlaf.util.UIScale;
 public class FlatSliderUI
 	extends BasicSliderUI
 {
-	protected int trackWidth;
-	protected int thumbWidth;
+	private int trackWidth;
+	private int thumbWidth;
 
-	protected Color trackColor;
-	protected Color thumbColor;
-	protected Color focusColor;
-	protected Color hoverColor;
-	protected Color disabledForeground;
+	private Color trackColor;
+	private Color thumbColor;
+	private Color focusColor;
+	private Color hoverColor;
+	private Color disabledForeground;
 
 	private MouseListener hoverListener;
 	private boolean hover;
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatSliderUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatSliderUI.java
@@ -61,14 +61,14 @@ import com.formdev.flatlaf.util.UIScale;
 public class FlatSliderUI
 	extends BasicSliderUI
 {
-	private int trackWidth;
-	private int thumbWidth;
+	protected int trackWidth;
+	protected int thumbWidth;
 
-	private Color trackColor;
-	private Color thumbColor;
-	private Color focusColor;
-	private Color hoverColor;
-	private Color disabledForeground;
+	protected Color trackColor;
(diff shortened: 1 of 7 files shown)


DevCharly:
Many thanks for your contribution.
Will try it soon...

> All other methods are copy/paste from BasicRangeSliderUI.

This is a license problem because the Jide license (GPL+classpath) is not compatible with the FlatLaf license (Apache).
To my understanding it is not legal to copy GPL+classpath licensed code into a Apache licensed project.

> It is not the best solution, but Jide have used the same approach for every look and feel (MetalRangeSliderUI, WindowsRangeSliderUi, ecc..).

I see that Jide uses copies of `BasicRangeSliderUI` in `MetalRangeSliderUI`, `WindowsRangeSliderUI` and `AquaRangeSliderUI`.
The reason is maybe that they are required to extend JDK classes like `WindowsSliderUI` or `AquaSliderUI` to use Windows/Mac specific painting. Don't know...

But in FlatLaf it is not that important to extend `FlatSliderUI` because painting done there is very simple.

It would be better that `FlatRangeSliderUI` extends Jide `BasicRangeSliderUI`.
Then you can copy needed code from `FlatSliderUI` to `FlatRangeSliderUI`.
This should avoid the license problem (and make the class much smaller).

What do you think?

basix86:
Ok, I will try!

basix86:
Hi, I have fixed the code and now FlatRangeSliderUI extends BasicRangeSliderUI.
I have a little bug now. 
When I change the theme, the foreground label color is not updated.

Have you any idea?

DevCharly:
> When I change the theme, the foreground label color is not updated

Found the reason.
In `JSlider` the labels are update in `JSlider.updateLabelUIs()` which is invoked from `JSlider.updateUI()`.
But `RangeSlider`, which extends `JSlider`, overrides `updateUI()` and does not invoke `updateLabelUIs()` or `super.updateUI()`.

Adding following code to `FlatRangeSliderUI` fixes the problem.

~~~java
	@Override
	public void installUI( JComponent c ) {
		super.installUI( c );

		// update label UIs, which is necessary because RangeSlider does not invoke JSlider.updateLabelUIs()
		updateLabelUIs( c );
	}

	@Override
	public void uninstallUI( JComponent c ) {
		// update label UIs also on uninstall to avoid light labels when switching
		// from dark FlatLaf theme to another Laf
		updateLabelUIs( c );

		super.uninstallUI( c );
	}

	protected void updateLabelUIs( JComponent c ) {
		Dictionary<?,?> labelTable = ((JSlider)c).getLabelTable();
		if( labelTable == null )
			return;

		Enumeration<?> e = labelTable.elements();
		while( e.hasMoreElements() ) {
			JComponent label = (JComponent) e.nextElement();
			label.updateUI();
			label.setSize( label.getPreferredSize() );
		}
	}
~~~

DevCharly:
Anti aliasing is not enabled for painting. Please enable with:

~~~java
	@Override
	public void paint( Graphics g, JComponent c ) {
		FlatUIUtils.setRenderingHints( (Graphics2D) g );

		super.paint( g, c );
	}
~~~

The "PaintTicks" and "PaintLabels" checkboxes should be selected by default.

Could you please undo all changes made to `FlatJideOssTest.java`.

basix86:
Thanks for your review.
I should have fixed the code

PR #214, opened 2020-11-25
Slider redesign

DevCharly:
This PR brings some visual changes to the slider (which makes it look beautiful IMHO)
and adds more possibilities to customize slider design.

This is the new default look (compared to old look):

image

Changes:
- the track is thinner
- the thumb 1px larger
- thumb (and left side of track) is always "blue", which makes it easier to recognize
- when focused, a "thick" border is painted around the thumb (inspired by material design)
- hover and pressed feedback on thumb
- height changed from 11 to 20 (without ticks and labels)
- customizable

Here are some examples to customize the look:

For Windows 10 style slider use:
~~~java
UIManager.put( "Slider.thumbSize", new Dimension( 8, 24 )  );
~~~
image

For macOS style slider use:
~~~java
UIManager.put( "Slider.thumbColor", Color.white );
UIManager.put( "Slider.thumbBorderColor", UIManager.getColor( "Component.borderColor" ) );
~~~
image

For thicker focus indicator (as in material design) use:
~~~java
UIManager.put( "Slider.focusWidth", 8 );
~~~
image

JIDE RangeSlider also uses new look and same customizing:

image


DevCharly:
Forgot to mention that the slider now returns a "useful" baseline to vertically align a label and the slider track when using a layout manager that supports baselines (`MigLayout` or `GridBagLayout`).

Notice how the label and the slider track are vertically aligned:

image

And this is what Windows Laf (and all other Lafs) do (not very useful IMHO):

image

As workaround (in Windows Laf), you can align the row to the top, which looks much better, but not perfect:
image

PR #222, opened 2020-12-05
AnimatedIcon

DevCharly:
This PR adds interface [AnimatedIcon](https://github.com/JFormDesigner/FlatLaf/blob/1293e2a07488977df27bbd3bfff68c88a9bded44/flatlaf-core/src/main/java/com/formdev/flatlaf/util/AnimatedIcon.java) (extends `javax.swing.Icon`)
that automatically animates painting on component value changes.

The cool thing is that the animation is done in the icon and it is not necessary to change anything in the UI delegates.
This is not yet used, but intended for future animates (issue #66).

The test application [FlatAnimatedIconTest](https://github.com/JFormDesigner/FlatLaf/blob/1293e2a07488977df27bbd3bfff68c88a9bded44/flatlaf-testing/src/main/java/com/formdev/flatlaf/testing/FlatAnimatedIconTest.java) includes (experimental) animated icons
for radio buttons, toggle switches and a minimal example:

animated-icon-200

Same as above but using 500ms duration to better see the animation:

animated-icon-500

Here is the code for a minimal example:

~~~java
private class AnimatedMinimalTestIcon
    implements AnimatedIcon
{
    @Override public int getIconWidth() { return 100; }
    @Override public int getIconHeight() { return 20; }

    @Override
    public void paintIconAnimated( Component c, Graphics g, int x, int y, float animatedValue ) {
        int w = getIconWidth();
        int h = getIconHeight();

        g.setColor( Color.red );
        g.drawRect( x, y, w - 1, h - 1 );
        g.fillRect( x, y, Math.round( w * animatedValue ), h );
    }

    @Override
    public float getValue( Component c ) {
        return ((AbstractButton)c).isSelected() ? 1 : 0;
    }
}

// sample usage
JCheckBox checkBox = new JCheckBox( "minimal" );
checkBox.setIcon( new AnimatedMinimalTestIcon() );
~~~

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/icons/FlatAnimatedIcon.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/icons/FlatAnimatedIcon.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 FormDev Software GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


Bios-Marcel:
Cool Idea and simple interface 👍

PR #229, opened 2020-12-21
TextField Placeholder now honors the right inset

MioBlanco:
Hey,

we added a custom icon to a textfield and used additional inset to do so. Since left and top inset is already being honored, we thought it'd make sense to do the same with the right inset.

However, there's two questions:

1. How do correctly format the source code (We used IDEA which should use .editorconfig, right?)
2. Should we clip the placeholder or use an ellipsis (Currently we are clipping it)

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTextFieldUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTextFieldUI.java
@@ -39,6 +39,7 @@ import javax.swing.text.Caret;
 import javax.swing.text.JTextComponent;
 import com.formdev.flatlaf.FlatClientProperties;
 import com.formdev.flatlaf.util.HiDPIUtils;
+import com.formdev.flatlaf.util.JavaCompatibility;
 
 /**
  * Provides the Flat LaF UI delegate for {@link javax.swing.JTextField}.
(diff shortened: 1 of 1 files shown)


DevCharly:
Thx, must admit that I did not try long placeholder text...

>     1. How do correctly format the source code (We used IDEA which should use .editorconfig, right?)

Yes, FlatLaf's [.editorconfig](https://github.com/JFormDesigner/FlatLaf/blob/master/.editorconfig) should define style used in FlatLaf.

But please do **not** reformat existing code.
This makes review hard and I do not merge such PRs.

When I add code to other projects, I simply look at existing code formatting and then write my code in that style.

>     2. Should we clip the placeholder or use an ellipsis (Currently we are clipping it)

Hmm, long text in text field is also simply clipped, but it is possible to scroll the text,
which is not possible for placeholder text...
So I think it would be better to use ellipsis (as in JLabel or JButton if they are too small).

Use method `com.formdev.flatlaf.util.JavaCompatibility.getClippedString()` for string clipping, which uses ellipsis.

Bios-Marcel:
I think it'd be good to have a CONTRIBUTION.md or a piece in the Readme that specifies the rules that we have to follow when making a PR.

So, in this case, what you are saying is, that the formatting is already correct, but just hasn't been applied in the past and therefore produces a big difference? Is there a way to avoid such a problem by specifying to only format edited lines or even disable formatting by default?

DevCharly:
> I think it'd be good to have a CONTRIBUTION.md or a piece in the Readme that specifies the rules that we have to follow when making a PR.

👍 

> So, in this case, what you are saying is, that the formatting is already correct, but just hasn't been applied in the past and therefore produces a big difference?

No, formatting of existing code is correct.

And on my machine, reformatting in IDEA with `.editorconfig` enabled does not reformat nearly the whole code (as in this PR).
There are only some differences. E.g. different line wrapping. But spaces are correct.

I do not use the formatter.
I write the code is so that it is easy to read.

Here is a comparison (from `FlatTextFieldUI`) of hand formatted code and IDEA formatted:

image

> Is there a way to avoid such a problem by specifying to only format edited lines or even disable formatting by default?

Don't know why IDEA reformatted the whole file on @MioBlanco computer.
On mine it does not reformat it.
It there some option to "reformat on save" in IDEA?

PR #237, opened 2021-01-14
JetBrains Runtime 11 has support for displaying the .AppleSystemUIFont font.

ingokegel:
This font should be used for UI elements since macOS 10.15.

See https://youtrack.jetbrains.com/issue/JBR-1915 for more information.

Other JREs, including JetBrains Runtime 8 do not handle kerning for that font correctly.


DevCharly:
Dankeschön 😃

ingokegel:
Ich habe zu danken! 

Btw, install4j 9 will include FlatLaf as the default LaF for installers. Of course I've added the credit for it in the install4j IDE and added some links to your site and this repository.

PR #239, opened 2021-01-14
Switched from sub-pixel to greyscale text anti-aliasing on macOS when running with a JetBrains JRE

ingokegel:
Sub-pixel anti-aliasing (VALUE_TEXT_ANTIALIAS_LCD_HRGB) causes font rendering with too much weight with a JetBrains JREs (both 8 and 11). This can be seen when comparing the text rendering of UI elements between IntelliJ IDEA and FlatLaf.

This commits aligns FlatLaf's behavior with IntelliJ IDEA which disables sub-pixel anti-aliasing on macOS for its IDE anti-aliasing setting and uses greyscale anti-aliasing by default (see com.intellij.ide.ui.AntialiasingType.canUseSubpixelAAForIDE).

Before:

JetBrains11_before

After:

JetBrains11_after

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatLaf.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatLaf.java
@@ -546,7 +546,12 @@ public abstract class FlatLaf
 	}
 
 	private void putAATextInfo( UIDefaults defaults ) {
-		if( SystemInfo.isJava_9_orLater ) {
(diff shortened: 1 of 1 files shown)


DevCharly:
Hmm, just realized that this change affects all platforms when using a JetBrains JRE.
Not only macOS.
Is this intended?

ingokegel:
You're right, I missed the SystemInfo.isMacOS condition. On Windows, subpixel text AA must be used with the JetBrains JRE. I've just submitted a new pull request.

Issue #228, opened 2020-12-17
FlatTableHeaderUI


thammoud:
Hello,

Thank you for such a wonderful library. We use a combination of JTable and Jide tables (SortableTable etc). Noticed that the header column borders do not have the borders painted as FlatLF installs an empty border and does the painting in the FlatTableHeaderUI.paintColumnBorders for normal JTables. Jide installs their own header UI's  and augment the painting using the DefaultTableHeaderRenderer. 

It feels that the standard way is to provide a column border that does all the painting so that any Swing table (Subclasses) will simply work. Just wanted to hear your thoughts on this. Perhaps I am missing something. Thank you.

Tarek

DevCharly:
> Thank you for such a wonderful library.

Thx 😃 

The reason for painting in `FlatTableHeaderUI` was to be able to paint the full scrollpane width
if JTable's autoResizeMode is set to OFF.

![image](https://user-images.githubusercontent.com/5604048/102899014-a1667280-446a-11eb-872e-5f017e59c166.png)

Here is what it looks like in Metal, which paints in column header borders:

![image](https://user-images.githubusercontent.com/5604048/102898996-9ca1be80-446a-11eb-81fc-cc96a0a09d43.png)

Anyway I agree with you that doing it the standard way would be better.
Probably a combination of using column border painting and painting in `FlatTableHeaderUI` could archive the same result.
Will try this...

DevCharly:
I've changed the painting in commit 4f2256f713c37310111e831398e648737abe6eed

It is now done the standard way in border `FlatTableHeaderBorder`

If you've sub-classed `FlatTableHeaderUI`, then this change may break it.
CC @ingokegel 

Please try the latest snapshot and report issues:
https://github.com/JFormDesigner/FlatLaf#snapshots

ingokegel:
Excellent and thanks for the heads up. It works fine if I just remove the call to `paintColumnBorders` in my derived classes.

PR #240, opened 2021-01-15
Made paintColumnBorders protected to help with implementing derived table header UIs

ingokegel:
To implement FlatLaf UIs for CellStyleTableHeaderUI and SortableTableHeaderUI from the Jide Grids library, access to the paintColumnBorders method is required.


DevCharly:
FYI the implementation of `FlatTableHeaderUI` may change for #228

PR #245, opened 2021-01-19
Added a per-tree wide selection setting

ingokegel:
Even when using wide selection by default, there are use cases where no selection background is desired, for example if the renderer shows the selected node in bold text without any selection background.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatClientProperties.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatClientProperties.java
@@ -670,6 +670,14 @@ public interface FlatClientProperties
 	 */
 	String TAB_BUTTON_SELECTED_BACKGROUND = "JToggleButton.tab.selectedBackground";
 
+	/**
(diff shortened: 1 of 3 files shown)


DevCharly:
Thanks, Ingo.

Good idea.
But using reflection to modify a private JRE field is a problem 
because this breaks functionality if access is denied by JRE,
which will be the default in Java 16 iI think
(see https://openjdk.java.net/jeps/396).

To avoid the use of reflection, I've implemented non-wide selection painting into `FlatTreeUI.paintRow()`
in the same way as done in `DefaultTreeCellRenderer.paint()` if `DefaultTreeCellRenderer.fillBackground` is `true`.
A renderer can paint over the selection background if necessary.

Have also added an additional client property to disable selection painting in `FlatTreeUI.paintRow()`.
Then the renderer is responsible for selection painting.

Please try this branch:
https://github.com/JFormDesigner/FlatLaf/tree/tree_wide_selection

ingokegel:
Thank you, this is an excellent solution and it's now very flexible!

PR #265, opened 2021-03-05
Make the module dependency on java.logging optional

ingokegel:
Currently, FlatLaf has the following module dependencies:

```
$ jdeps --list-deps --multi-release 9 flatlaf-1.0.jar
   java.base
   java.desktop
   java.logging
```
This commit makes the java.logging dependency optional and hides logging behind a facade that falls back to printing to stderr if the java.logging module is not available.

To test, create a reduced JRE with a command like

```
$ jdk-15/bin/jlink.exe --module-path jdk-15/jmods  \
    --add-modules java.desktop  \
    --add-modules java.instrument  \
    --output jre-15-desktop-only
```

(adding java.instrument, so the FlatLafDemo main class can be started from IntelliJ IDEA)


DevCharly:
Thanks, Ingo.

Tried it, but it does not reduce JRE size that much: only 230 KB of 71,4 MB (OpenJDK 15.0.2).

Anyway I'm open to merge it with some changes:

- since class `LoggingFacade` needs to be public, I think it better fits into the `com.formdev.flatlaf.util` package
- please add an Apache license header to `LoggingFacade.java` (copyright is yours)
- remove `logSevere( Throwable t )` and `logSevere( String message )` and always use `logSevere(String message, Throwable t)`
- to avoid NPEs when printing to stderr, check `message` and `t` for `null`

ingokegel:
Done.  I would not like to claim a separate copyright for my small pull request, so I've added your header.

The reason for this pull requests is that in install4j JRE bundle creation is an important topic and some people complained that java.util.logging was added through the FlatLaf dependency.

Another possibility would be to use the multi-release feature of Java 9 and use the new system logger in java.base:

https://docs.oracle.com/javase/9/docs/api/java/lang/System.Logger.html

Then there would be no exception handling in LoggingFacade by default and it would be functionally better on Java 9+. What do you think?

DevCharly:
Didn't know that there is new logger in Java 9.
Sounds like a very good idea to use it 👍 

Since FlatLaf already is a MRJ, it is easy to add a new Java 9+ class to `flatlaf-core/src/main/java9`

ingokegel:
OK, please have a look. I've tested that it uses java.util.logging with JDK 8 and System.Logger with Java 11. The output on the console is the same.

This is the first time an API is used that is not in recent Java 1.8 JDKs (MultiResolutionImage is actually present there). So in the IDE, you now have to use a Java 9+ JDK as the project JDK whereas previously a 1.8 JDK worked as well.

DevCharly:
Thanks Ingo. Works perfect.

Was able to eliminate optional module `java.logging` from `module-info.java` by fixing the Gradle task `:flatlaf-core:compileModuleInfoJava`. And this was a good idea. Had some java.util logging in another PR, which I just merged. Without removing `requires static java.logging`, I maybe had not noticed this...

Issue #170, opened 2020-09-07
Windows: Snap to left/right half of screen not working


erl987:
Hi,

if you take the FlatLaf demo app (or some real-world application using FlatLaf), it is not possible to *snap the application to the left or right half of the screen* (using `Windows Key` + `Left Arrow`, etc.).

This would be a great enhancement to consider in the further development of FlatLaf.

Best regards,
Ralf

stanio:
I've noticed this happens when painting windows decorations is on (which is by default for the demo).  If one unchecks Options / Windows decorations, snapping gets working.

DevCharly:
Yes, happens only if FlatLaf **custom window decorations** are enabled.

This is a feature of Windows 10, which is unfortunately not available for undecorated frames in Java (used for custom window decorations). So even if I implement e.g. `Windows+Left` keys in FlatLaf and resize the window to fill the left half of the screen, this would not show the other open windows as icons on the right half of the screen...

There is an article that solves this by using Java Native Access (JNA):
https://medium.com/swlh/customizing-the-title-bar-of-an-application-window-50a4ac3ed27e
But this is no option for core FlatLaf because the necessary library dependencies are 4-5 MB.
Maybe this can be done in a FlatLaf subproject...

An option that already works is to **use JetBrains Runtime 11** to run own application:
https://confluence.jetbrains.com/display/JBR/JetBrains+Runtime
JetBrains Runtime (JBR) is based on OpenJDK and powers IntelliJ IDEA and other JetBrains IDEs.
JBR supports native Windows features for undecorated frames and FlatLaf supports JBR.
So even if FlatLaf custom window decorations are enabled, application window behaves like native window.

Hope that JBR custom window decorations are ported to OpenJDK in the future...

erl987:
Thanks, we go for switching off the custom window decorations - at least for now. Using Java 11 is not really an option for us.

Cheers,
Ralf

DevCharly:
There is now a fix available in PR #267 
Would be great if you could try it out in your app.
Thx.

erl987:
Great! We will test this as soon as possible!

DevCharly:
now fixed in main branch

erl987:
I can confirm that the application is now working as expected with FlatLaf 1.1.

Thanks for the great work!

PR #267, opened 2021-03-10
Native window decorations for Windows 10 (using JNI)

DevCharly:
This PR is the successor of PR #262 and implements native window decorations for Windows 10 (64bit only) using JNI. This avoids the 3MB dependency on JNA. Also `flatlaf-native-jna-1.1-SNAPSHOT.jar` is no longer needed. A small 9kB DLL is now contained in `flatlaf.jar`. This DLL is extracted at startup to folder `$TEMP/flatlaf.temp`. Since it is not possible to delete it when the application exits (because Windows has locked it), unused temporary DLL is deleted on next startup.

This PR enables dark frame/dialog title bar and embedded menu bar with all JREs while still having:
- native Window 10 border drop shadows
- native Window 10 resize functionality
- [Windows 10 snapping](https://support.microsoft.com/en-us/windows/snap-your-windows-885a9b1e-a983-a3b1-16cd-c531795e6241) functionality
- native Windows 10 system window menu when right-clicking on title bar or left-clicking on application icon

image

image

## Try it

To try it out, either build from source or download the build artifacts ZIP from GitHub Actions 
(at the bottom of that page):
https://github.com/JFormDesigner/FlatLaf/actions/runs/642361750

The ZIP contains the demo `flatlaf-demo-1.1-SNAPSHOT.jar` (contains all dependencies).

To test in own applications, add `flatlaf-1.1-SNAPSHOT.jar` to the classpath.

## Enable/disable

This feature is **enabled by default** on Windows 10 64-bit.

If you don't like/want it, you can disable it with:
~~~java
UIManager.put( "TitlePane.useWindowDecorations", false );
~~~

It is also possible to disable only the embedded menu bar (and keep the dark title pane) with:
~~~java
UIManager.put( "TitlePane.menuBarEmbedded", false );
~~~

It is also possible to disable this on command line with following VM options:
~~~
-Dflatlaf.useWindowDecorations=false
-Dflatlaf.menuBarEmbedded=false
~~~

If you have following code in your app, you **can remove it** (no longer necessary):
~~~java
// enable window decorations
JFrame.setDefaultLookAndFeelDecorated( true );
JDialog.setDefaultLookAndFeelDecorated( true );
~~~

## JetBrains Runtime (JRE)

If you're using JetBrains Runtime 11 to run your app, this feature is **not used**.
Instead the native window decorations feature provided by the JetBrains Runtime is used (as in FlatLaf 1.0).

fixes issue #170


DevCharly:
This feature is now enabled by default (see updated main post)

basix86:
Very nice and complex job.
Congratulations!

PR #276, opened 2021-03-22
Added ButtonType.borderLess

ingokegel:
To create buttons that look like toolbar buttons but have a focus indicator.

This behavior can be achieved with JideButton, but it would be preferable to use FlatButton instead.


DevCharly:
Thanks. Tested. LGTM.

Any reason that you used two words `border less` (or `borderLess`)?
Isn't it a single word `borderless`?

ingokegel:
You're right, it's borderless, I just renamed all occurrences

PR #278, opened 2021-03-23
Add a setter for the native provider

ingokegel:
This makes it possible to support situations where the extraction of a DLL at runtime is not possible.

Rationale:

Extracting DLLs to the temporary directory will trigger some anti-virus tools and can lead to quarantined binaries.
This is why in install4j we use the RegisterNatives JNI function to provide native code from the installer executable without loading an additional library. For the native code in FlatLaf, we would like to do the same.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatNativeWindowBorder.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatNativeWindowBorder.java
@@ -247,14 +247,17 @@ public class FlatNativeWindowBorder
 /*
 			Class<?> cls = Class.forName( "com.formdev.flatlaf.natives.jna.windows.FlatWindowsNativeWindowBorder" );
 			Method m = cls.getMethod( "getInstance" );
-			nativeProvider = (Provider) m.invoke( null );
+			setNativeProvider( (Provider) m.invoke( null ) );
 */
-			nativeProvider = FlatWindowsNativeWindowBorder.getInstance();
-
-			supported = (nativeProvider != null);
+			setNativeProvider( FlatWindowsNativeWindowBorder.getInstance() );
 		} catch( Exception ex ) {
 			// ignore
 		}
+	}
+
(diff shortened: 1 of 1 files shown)


ingokegel:
Btw, congrats to the native Window decorations for all JREs, this is an awesome feature!

DevCharly:
Thanks, merged.
Have added an additional check to `setNativeProvider()` to prevent multiple invocations.

> Btw, congrats to the native Window decorations for all JREs, this is an awesome feature!

Thanks. I'm very happy about this feature. It makes it so much better...

> Extracting DLLs to the temporary directory will trigger some anti-virus tools and can lead to quarantined binaries.

That's interesting. I wonder what happens in this case.
Does `System.load()` simply fail with an exception?
Or does the application crash?
Is there anything in FlatLaf that needs improvement for this?

> This is why in install4j we use the RegisterNatives JNI function to provide native code from the installer executable without loading an additional library. For the native code in FlatLaf, we would like to do the same.

Interesting. I thought you're always using JetBrains Runtime for the installers.

I wonder how you use FlatLaf native code. Can you simply use the FlatLaf DLLs?

ingokegel:
Thank you!

> Does System.load() simply fail with an exception?
>Or does the application crash?

We have no reports about native crashes, but about UnsatisfiedLinkErrors when loading a library when an anti-virus tool immediately quarantines the DLL after it has been extracted. So it would probably be good to protect against this condition. I also think that offering a mode where the FlatLaf DLLs can be placed next to the JAR file (or into a configurable path) would be a good idea.

>Interesting. I thought you're always using JetBrains Runtime for the installers.

For install4j projects, you can select the JDK provider from AdoptOpenJDK/Corretto/JetBrains/Liberica/Zulu and AdoptOpenJDK is the default. We use the JetBrains runtime for our own products.

>I wonder how you use FlatLaf native code. Can you simply use the FlatLaf DLLs?

For the installer, we will compile the code into the installer executable and write a FlatLaf native provider that resolves these calls with the registerNatives JNI function.

PR #297, opened 2021-03-31
Add issue template for reports in order to reduce communication overhead

Bios-Marcel:
Thought it might be useful to avoid the basic ping pong question game at the start of each issue.


Bios-Marcel:
I'll just be closing this. I'd still recommend some kind of issue template though :)

PR #298, opened 2021-03-31
Fix selected states for native window border related menu items

Bios-Marcel:
The menu items for custom window decorations and embeded menu bar aren't selected anymore if the feature isn't supported.
On top of that, there's now a tooltip indicating that these aren't supported.

Regarding the previous condition for enabling the item:

```java
supportsWindowDecorations && !JBRCustomDecorations.isSupported()
```

I deem it very confusing that we set the item to `enabled` if `JBRCustomDecorations` isn't supported.
Would you mind explaining this? I'd like to add an explanatory comment in the code here.

### DIFF
--- a/flatlaf-demo/src/main/java/com/formdev/flatlaf/demo/DemoFrame.java
+++ b/flatlaf-demo/src/main/java/com/formdev/flatlaf/demo/DemoFrame.java
@@ -750,8 +750,16 @@ class DemoFrame
 
 		boolean supportsWindowDecorations = UIManager.getLookAndFeel()
 			.getSupportsWindowDecorations() || FlatNativeWindowBorder.isSupported();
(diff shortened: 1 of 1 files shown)


DevCharly:
> I deem it very confusing that we set the item to enabled if JBRCustomDecorations isn't supported.

Yes, that looks confusing, but is correct. If running in a JetBrains VM, then the window decorations provided by the JetBrains VM are used and it is **not possible to disable them** at runtime. Therefore the "Window decorations" menu item is disabled if `JBRCustomDecorations.isSupported()` returns `true`.

Bios-Marcel:
Thanks for explaining. Will add an explanatory comment when I am home :)
Unless you want to do to yourself, whatever is easier for you.

Bios-Marcel:
Done :)

Issue #340, opened 2021-06-11
[Question] Does FlatLaf Support CSS Styling in any way?


orange451:
I am trying to develop an OpenGL application within Java, and need to pick a UI solution. I spent a lot of time in the past working on my own UI solution that supported some basic CSS styling. I was wondering if FlatLaf supports anything like this? I fear my solution is too slow for large applications, so I want to try other libraries.

DevCharly:
There is some initial work going on in PR #341 to support CSS styling on components (using client properties), but this does not allow using a CSS file.

If you're looking for a way to "style" all components, you can do this in FlatLaf properties files:

- https://github.com/JFormDesigner/FlatLaf/blob/main/flatlaf-core/src/main/resources/com/formdev/flatlaf/FlatLaf.properties
- https://github.com/JFormDesigner/FlatLaf/blob/main/flatlaf-core/src/main/resources/com/formdev/flatlaf/FlatLightLaf.properties
- https://www.formdev.com/flatlaf/properties-files/
- https://www.formdev.com/flatlaf/how-to-customize/

Issue #117, opened 2020-06-24
API to override L&F provided attributes (colors, insets, etc.)


Chrriis:
Hi,

I was experimenting having a menu bar with dark colors (just the menu bar, not its menus) and I realized that I cannot set all colors:
![CustomMenuBar](https://user-images.githubusercontent.com/1174202/85551063-e75c8e00-b621-11ea-9b6b-1d02e8d6d0ad.png)

I understand that the purpose of the look and feel is to define all colors for us. Still, is there a way to override colors for a given component? If not, would such API make sense?

Possible candidates off the top of my head:
- `menu.putClientProperty("FlatLaf.attribute.MenuItem.underlineSelectionBackground", Color.BLUE);`
- subclassing JMenu to provide this information somehow.
- changing the UI delegate (though this is a bit too much in my view).
- have the notion of a theme variant per component: e.g.: `menu.putClientProperty("FlatLaf.variant", "SpecialBar")`, which would load the properties from the look and feel named "SpecialBar.MenuItem.underlineSelectionBackground" with fallback to main variant for properties that are not defined.
- add "MenuBar.inverseColors" as a standard capability. It is what I wanted to achieve, but it is not uncommon these days for toolbars/menubars.

My test case:
[MenuBarWithStyle.zip](https://github.com/JFormDesigner/FlatLaf/files/4825177/MenuBarWithStyle.zip)

DevCharly:
The menu hover colors are now fixed in latest commit.
They are derived at runtime from the background color.

To make the menu bar dark use:
~~~java
UIManager.put( "MenuBar.background", Color.DARK_GRAY );
UIManager.put( "MenuBar.foreground", Color.WHITE );
~~~
or
~~~java
menuBar.setBackground( Color.DARK_GRAY );
menuBar.setForeground( Color.WHITE );
~~~

Changing the foreground color of the top-level `JMenus` is not necessary. 
They now use the foreground of the `JMenuBar`.

Chrriis:
Neat! 👍 

If I use the underline style for menus, is there a way to set the underline color of the menubar?

DevCharly:
Menu underline color can be changed with (for all menu items):
~~~java
UIManager.put( "MenuItem.underlineSelectionColor", Color.RED );
~~~

If you want it only for the menu bar, there is a trick that works:
~~~java
UIManager.put( "MenuItem.underlineSelectionColor", Color.RED );

// create JMenus for menubar here
[...]

PR #341, opened 2021-06-15
Styling individual components

DevCharly:
Currently most FlatLaf styling capabilities are limited to change **all components** at once.
This PR is an attempt to better support styling **individual** components.

Some "styling" properties are available via Swing API (background, foreground, border, font, etc) for individual components,
but many FlatLaf UI defaults can be defined only for all components in FlatLaf properties files (https://www.formdev.com/flatlaf/properties-files/) or with `UIManager.put(key, value)`.

The idea is to use a single client property to modify multiple style properties of a component.
This is similar to what can be done in JavaFX with `Node.setStyle(String style)`.

## CSS syntax

Therefore I've chosen CSS syntax: `key1: value1; key2: value2; ...`

~~~java
mySlider.putClientProperty( "FlatLaf.style", "trackWidth: 2" );
~~~

The key is the same as used in UI defaults, but without component type prefix.
E.g. for UI default `Slider.thumbSize` use key `thumbSize`.
The syntax of the value is the same as used in FlatLaf properties files
(https://www.formdev.com/flatlaf/properties-files/),
but some features are not supported (e.g. variables).

Color functions and references are supported. E.g::

~~~java
mySlider.putClientProperty( "FlatLaf.style", "thumbColor: lighten(#f00,20%)" );
mySlider.putClientProperty( "FlatLaf.style", "trackColor: lighten($TextField.foreground,20%)" );
mySlider.putClientProperty( "FlatLaf.style", "tickColor: $TextField.foreground" );
~~~

`$TextField.foreground` is replaced with the value returned by `UIManager.get( "$TextField.foreground" )`.

## Map

Instead of the CSS string, it is also possible to assign a `java.util.Map` to the client property. 
In this case the values are not parsed from a string. They must be binary:

~~~java
Map<String, Object> map = new HashMap<>();
map.put( "trackWidth", 2 );
map.put( "thumbColor", new Color( 0xf0f0f0 ) );
mySlider.putClientProperty( "FlatLaf.style", map );
~~~

## Supported components

- `JButton` and `JToggleButton`
- `JCheckBox` and `JRadioButton` (including icon)
- `JComboBox` and `JSpinner`
- `JInternalFrame`
- `JLabel`
- `JList`, `JTable`, `JTableHeader` and `JTree`
- `JMenuBar`, `JMenu`, `JMenuItem`, `JCheckBoxMenuItem`, `JRadioButtonMenuItem` and `JPopupMenu`
- `JProgressBar`
- `JSeparator` and `JPopupMenu.Separator`
- `JScrollPane` and `JScrollBar`
- `JSlider`
- `JSplitPane`
- `JTabbedPane`
- `JTextField`, `JFormattedTextField` and `JPasswordField`
- `JTextArea`, `JTextPane` and `JEditorPane`
- `JToolBar` and `JToolBar.Separator`

## Possible "styles"

For possible "styles" take a look at:
https://github.com/JFormDesigner/FlatLaf/blob/styling/flatlaf-core/src/test/java/com/formdev/flatlaf/ui/TestFlatStyling.java

It is also possible to "style" any JavaBean property (that has public getter and setter methods).
For CSS strings this works only for property types that can be parsed from a string.

~~~java
mySlider.putClientProperty( "FlatLaf.style", "maximum: 200; value: 100; majorTickSpacing: 50; paintLabels: true" );
~~~

## Platform specific and light/dark theme specific styles

If a key is preceded by `[win]`, `[mac]` or `[linux]`, then its value is only used on the given platform.
If a key is preceded by `[light]` or `[dark]`, then its value is only used if the current FlatLaf theme is light or dark.

~~~java
mySlider.putClientProperty( "FlatLaf.style",
    "[mac]trackWidth: 4; [light]thumbColor: #000; [dark]thumbColor: #fff" );
~~~

This is also supported in FlatLaf properties files:
https://www.formdev.com/flatlaf/properties-files/#platform_specific

## JSlider examples

Windows 10 style slider:
~~~java
mySlider.putClientProperty( "FlatLaf.style", "thumbSize: 8,24" );
~~~
image

macOS style slider:
~~~java
mySlider.putClientProperty( "FlatLaf.style", "thumbColor: #fff; thumbBorderColor: #c4c4c4" );
~~~
image

Thicker focus indicator (as in material design):
~~~java
mySlider.putClientProperty( "FlatLaf.style", "focusWidth: 8" );
~~~
image

CC @Chrriis (issue #117); @orange451 (issue #340)


Chrriis (changes requested):
I looked at the code and this is good!
I would only suggest one change: I think all FlatLaf-specific client properties should indicate so: instead of `"JComponent.style"`, I think it would be better to use `"FlatLaf.style"`.

Chrriis:
Something crossed my mind: how would you defined more complex objects, e.g. an icon?
One possibility would be to use indirections:
```Java
myTree.putClientProperty("FlatLaf.style", "expandedIcon: chrriis.expandedIcon; collapsedIcon: chrriis.collapsedIcon");
myTree.putClientProperty("chrriis.expandedIcon", myExpandedIcon);
myTree.putClientProperty("chrriis.collapsedIcon", myCollapsedIcon);
```

DevCharly:
> Something crossed my mind: how would you defined more complex objects, e.g. an icon?

Good question. 

In the meantime it is possible to pass a `Map<String, Object>` to the client property (instead of a CSS string).
Then following is possible:
 
```java
Map<String, Object> map = new HashMap<>();
map.put( "expandedIcon", myExpandedIcon );
map.put( "collapsedIcon", myCollapsedIcon );
myTree.putClientProperty( "FlatLaf.style", map );
```

References to UI default values are already possible in CSS string syntax.
E.g. following invokes `UIManager.get( "MySlider.someColor" )` to get the value for `thumbColor`.
```java
mySlider.putClientProperty( "JComponent.style", "thumbColor: $MySlider.someColor" );
```

Maybe it makes sense to first look into client properties and then into `UIManager`...

Chrriis:
> In the meantime it is possible to pass a Map<String, Object> to the client property (instead of a CSS string).

I will definitely use a map instead of the CSS syntax for my use cases!

One thing I am not sure about is that I cannot simply copy/paste and tweak the definitions from UIManager default values to the client properties map because the keys are different: I have to remove all the prefixes. That being said, styling a specific component is for special cases, so having to remove the prefixes is not a big deal.

flatlaf-core/src/main/java/com/formdev/flatlaf/UIDefaultsLoader.java
```diff
@@ -351,14 +351,15 @@ else if( value.startsWith( "\"" ) && value.endsWith( "\"" ) ) {
 				valueType = ValueType.STRING;
 			else if( key.endsWith( "Color" ) ||
 				(key.endsWith( "ground" ) &&
-				 (key.endsWith( ".background" ) || key.endsWith( "Background" ) ||
-				  key.endsWith( ".foreground" ) || key.endsWith( "Foreground" ))) )
+				 (key.endsWith( ".background" ) || key.endsWith( "Background" ) || key.equals( "background" ) ||
+				  key.endsWith( ".foreground" ) || key.endsWith( "Foreground" ) || key.equals( "foreground" ))) )
```
Bios-Marcel, line 355: Wouldn't `endsWithIgnorecase( "foreground"/"background")` suffice and also be faster?
Same for the other instances where you have this pattern?

DevCharly: Method `endsWithIgnorecase()` does not exist 😄 

And following line avoids that this is executed in most cases:

https://github.com/JFormDesigner/FlatLaf/blob/82192bef913254755f43752a5e74c7f4a65b3957/flatlaf-core/src/main/java/com/formdev/flatlaf/UIDefaultsLoader.java#L353

`equals( "background" )` is very fast because it first compares the string lengths.
`startsWith()` and `endsWith()` are also very fast because they return immediately when a character is not equal.

Bios-Marcel: That's absolutely true :D

PR #343, opened 2021-06-20
New tabbed pane active tab border painting style

xDUDSSx:
This pull request adds a new option to tabbed panes that changes how their active tab and content separator is painted.
It makes the content separator wrap around the active tab much like many typical tabbed pane designs (like metal laf).

newTabsCrop

### Why?
- It looks good and it's a widespread design that I've personally been missing in flatlaf. Aside from being present in most UI frameworks on desktop, it can also be found in flat web UI's made with react for example.
- Generally, this design makes the active tab (and whole tabbed pane) pop out / visible a lot more. 
I feel like the current flatlaf material design-ish tabs tend to be hard to find in certain situations/UIs. After all the only indicator of a tabbed pane at first sight (aside from content of the tabs themselves) is just the tiny selection indicator. That's fine for the most part and looks very clean, but in a more complex UI, especially when the tabbed pane is surrounded with other components, it can be hard to find. There is no visual indicator that would separate the top of the tabbed pane and the content above. This design makes the separator bend upwards around the active tab and outline the top of the tabbed pane where the active tab is.
- If the tab and content background are the same they visually merge together which is quite nice and, well, like a tab.

It can be enabled / disabled with the key `activeTabBorder` exactly like `showTabSeparators` is used.

So for example
```
tabbedPane.putClientProperty("JTabbedPane.activeTabBorder", true);
```
or
```
UIManager.put("TabbedPane.activeTabBorder", true);
```

It moves the selection indicator to the top and paints two lines on the sides of the active tab. The tab separator is drawn in parts to leave a gap where the active tab is. In reality, there is no gap since it's filled with a line of the active tab's background color to prevent an ugly line when the active tab is focused or hovered over. This was a bit of a pain since I had to add an extra repaint method for just the separator and handle repaint issues with focus. It does seem to work quite well though.

I've tested everything I thought of in the demo.
- Works with scrolling
- Works with all tab placements
- Works with trailing/leading components
- Works with the full border
- Works with tab separators
- Works with thicc separator height
- Works with different selection indicator heights

I haven't created any "rigorous" tests so I could have easily missed something.

### Example from React
yee

I did add a demo to the demo. I did not edit the changelogs or anything.
Hope you like it.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatClientProperties.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatClientProperties.java
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-
 package com.formdev.flatlaf;
+
 import java.awt.Color;
 import java.util.Objects;
 import javax.swing.JComponent;
@@ -325,21 +325,41 @@ public interface FlatClientProperties
 	//---- JTabbedPane --------------------------------------------------------
 
 	/**
+	 * Specifies type of the selected tab.
+	 * <p>
+	 * <strong>Component</strong> {@link javax.swing.JTabbedPane}<br>
+	 * <strong>Value type</strong> {@link java.lang.String}<br>
+	 * <strong>Allowed Values</strong>
+	 *     {@link #TABBED_PANE_TAB_TYPE_UNDERLINED} or
+	 *     {@link #TABBED_PANE_TAB_TYPE_CARD}
+	 *
+	 * @since 2
+	 */
+	String TABBED_PANE_TAB_TYPE = "JTabbedPane.tabType";
+
+	/**
+	 * Paint the selected tab underlined.
+	 *
+	 * @see #TABBED_PANE_TAB_TYPE
+	 * @since 2
+	 */
+	String TABBED_PANE_TAB_TYPE_UNDERLINED = "underlined";
+
+	/**
+	 * Paint the selected tab as card.
+	 *
+	 * @see #TABBED_PANE_TAB_TYPE
+	 * @since 2
+	 */
+	String TABBED_PANE_TAB_TYPE_CARD = "card";
+
+	/**
 	 * Specifies whether separators are shown between tabs.
 	 * <p>
 	 * <strong>Component</strong> {@link javax.swing.JTabbedPane}<br>
 	 * <strong>Value type</strong> {@link java.lang.Boolean}
 	 */
 	String TABBED_PANE_SHOW_TAB_SEPARATORS = "JTabbedPane.showTabSeparators";
-
-	/**
-	 * Specifies whether a border is painted around the active tab.
-	 * This also changes position of the active tab indicator.
-	 * <p>
-	 * <strong>Component</strong> {@link javax.swing.JTabbedPane}<br>
-	 * <strong>Value type</strong> {@link java.lang.Boolean}
-	 */
-	String TABBED_PANE_ACTIVE_TAB_BORDER = "JTabbedPane.activeTabBorder";
 
 	/**
 	 * Specifies whether the separator between tabs area and content area should be shown.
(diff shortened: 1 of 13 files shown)


DevCharly:
Nice 👍 👍  I'll try it...

xDUDSSx:
Added a commit to make the active tab borders the same width as the separator.
thickSeparator

The active tab also paints an upper border if the tab selection height is 0.
noSelectionIndicator

Also, I've realized that I haven't tested this with UI scaling, but I checked that all dimensions get scaled.

DevCharly:
@xDUDSSx sorry for the delay.
I've reviewed/tested the PR.
Great work!

It worked perfecty, but anyway made some changes to make the code shorter:

- moved focus listener to class `Handler`
- instead of repainting content border in `repaintContentBorder()`, increase size of repaint region in `repaintTab()` to include part of content border
- simplified `paintContentBorder()` by using `getTabBounds()` and `Rectangle2D.intersect()` to compute gap rectangle
- slightly make code smaller
- minor formatting, renaming, ...

The only visual change is that antialiasing is used to paint tab border (and tab separators), which is only visible if using Java 9+ and scaling (e.g. 125%).

The focus listener is indeed necessary because the `BasicTabbedPaneUI` focus listener repaints only the tab, but not the separator below the tab.

What I'm not happy about is the name of the key `TabbedPane.activeTabBorder`. 
The suffix `Border` is usually used to specify borders.
And `activeTabBorder` sounds like defining the border for the active tab.

I know that there are two other keys that end with `Border` that are not borders, but `TabbedPane.tabsOverlapBorder` is defined in Swing's `BasicTabbedPaneUI` and `TabbedPane.hasFullBorder` is used for compatibility with IntelliJ `DarculaTabbedPaneUI`: https://github.com/JetBrains/intellij-community/blob/0e2aa4030ee763c9b0c828f0b5119f4cdcc66f35/platform/platform-impl/src/com/intellij/ide/ui/laf/darcula/ui/DarculaTabbedPaneUI.java#L120

Maybe a property `TabbedPane.tabStyle` or `TabbedPane.tabType` (similar to `TabbedPane.arrowType` or client property `JButton.buttonType`) of type `String` would be better for future extensibility with more tab styles. Values could be e.g. `"underline"`´ and `"borderOnSelectedTab"` (or similar).

xDUDSSx:
@DevCharly Thanks for reviewing. Glad it somewhat worked.

The key name can certainly be changed, I couldn't really figure out what to call it exactly so I just threw that in as a key at the time.

I think having a key like `tabStyle` would be ideal. I'm not so sure what to call this style though 😄 and I can't seem to find any specific references to it anywhere (or rather to its name). For the current style `underline` sounds good. The new style could be `wrapAround` or something. `borderOnSelectedTab` sounds a bit confusing.

DevCharly:
Naming is hard...

Found a similar style named "container" here https://www.carbondesignsystem.com/components/tabs/usage/#live-demo

image
image

Here is one named "cards", which has a border on all tabs:
https://element-plus.org/#/en-US/component/tabs#card-style

image

Here is something useful https://chakra-ui.com/docs/disclosure/tabs#tab-variants-and-color

variant "line":
image

variant "enclosed":
image

variant "enclosed-colored":
image

variant "soft-rounded" (or "solid-rounded"):
image

xDUDSSx:
Indeed. You seem to be better at googling than me at least.

I'd vote for enclosed or card. Enclosed to me sounds more representative of the style itself. A "card" sounds more like a look achieved with enclosed style and full height separators.

Also, the rounded style could be another candidate for a third tab style 😄.

DevCharly:
Finally, I've merged it 😃 

Have replaced `activeTabBorder` with `tabType`. Allowed values are `"underlined"` and `"card"`.
(did not want use `tabStyle` to avoid confusion with general styling (PR #341) coming in v2)

Have updated the initial post with the new client property and UI default key.

Thank you very much 👍

PR #388, opened 2021-09-25
Styling: basic support for "classes" (similar to CSS classes)

DevCharly:
PR #341 added basic styling for components. E.g.
~~~java
mySlider.putClientProperty( "FlatLaf.style", "thumbColor: #f00" );
~~~
which is comparable to what can be done in HTML:
~~~html
<p style="background: #f00">
~~~

This PR is the next step and adds support for "style classes". E.g.
~~~java
mySlider.putClientProperty( "FlatLaf.styleClass", "myclass" );
or
mySlider.putClientProperty( "FlatLaf.styleClass", "myclass1 myclass2" );
~~~
which is comparable to what can be done in HTML:
~~~html
<p class="myclass">
or
<p class="myclass1 myclass2">
~~~

This allows defining style rules at a single place and use them everywhere.

The style rules must be defined in UI defaults either as strings (in CSS syntax) or as `Map<String, Object>` (with binary values).
The key must be in syntax: `[style]type.styleClass`, where the type is optional.
E.g. in FlatLaf properties file:
~~~
[style]Button.primary = borderColor: #08f; background: #08f; foreground: #fff
[style].secondary = borderColor: #0f8; background: #0f8
~~~
or in Java code:
~~~java
UIManager.put( "[style]Button.primary", "borderColor: #08f; background: #08f; foreground: #fff" );
UIManager.put( "[style].secondary", "borderColor: #0f8; background: #0f8" );
~~~
The rule `Button.primary` can be applied to buttons only.
The rule `.secondary` can be applied to any component.

To have similar behavior as in CSS, first the rule without type is applied, then the rule with type.
E.g. setting style class to `foo` on a `JButton` uses rules from UI default keys `[style].foo` and `[style]Button.foo`.

CC @Chrriis (issue #117); @orange451 (issue #340)

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatClientProperties.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatClientProperties.java
@@ -146,8 +146,8 @@ public interface FlatClientProperties
 	String STYLE = "FlatLaf.style";
 
 	/**
-	 * Specifies the style class(es) of a component as String
-	 * or as {@link java.util.List}&lt;String&gt;.
+	 * Specifies the style class(es) of a component as String (single class or multiple classes separated by space characters)
+	 * or as {@code String[]} or {@link java.util.List}&lt;String&gt; (multiple classes).
 	 * <p>
 	 * The style rules must be defined in UI defaults either as strings (in CSS syntax)
 	 * or as {@link java.util.Map}&lt;String, Object&gt; (with binary values).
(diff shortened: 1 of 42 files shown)


DevCharly:
I wonder whether it would be better to use a comma separated string for multiple classes instead of a string array?
~~~java
mySlider.putClientProperty( "FlatLaf.styleClass", "myclass1, myclass2" );
~~~
instead of
~~~java
mySlider.putClientProperty( "FlatLaf.styleClass", new String[] { "myclass1", "myclass2" } );
~~~
which is comparable to what is done in HTML:
~~~html
<p class="myclass1, myclass2">
~~~

Any opinions?

ebourg:
CSS classes in HTML are actually specified without a comma, so the syntax could be simplified to:

    mySlider.putClientProperty( "FlatLaf.styleClass", "myclass1 myclass2" );

DevCharly:
@ebourg many thanks for pointing out my mistake
@Chrriis yes

PR #415, opened 2021-11-03
MenuItem: paint the selected icon when the item is selected

ebourg:
I'd like to suggest this change to paint a different icon when a menu item is highlighted. The idea is to provide an alternative icon that has a better contrast with the selection background. This behavior can be seen for example on some menus of IntelliJ where the icons are painted with a different shade of gray when an item is selected:

**Open icon not selected (color #6E6E6E)**

**Open icon selected (color #AFB1B3)**

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatMenuItemRenderer.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatMenuItemRenderer.java
@@ -35,6 +35,8 @@ import javax.swing.Icon;
 import javax.swing.JMenu;
 import javax.swing.JMenuItem;
 import javax.swing.KeyStroke;
+import javax.swing.MenuElement;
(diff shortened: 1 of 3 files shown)


DevCharly:
Good idea 👍 Thanks.

But better use `isArmedOrSelected( menuItem )` instead of `MenuSelectionManager`.

DevCharly:
... and move it after the `if( menuItem.getModel().isPressed() && menuItem.isArmed() ) { ... }` block. Otherwise, this block could never be reached...

https://github.com/JFormDesigner/FlatLaf/blob/6b880af447c8f903c76e2f7bc4e21fabd409f403/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatMenuItemRenderer.java#L484-L488

At the moment I'm not sure what this block is good for. Have a breakpoint in the block, but it is never reached. Seems that menu items are not "pressed". Strange...

Issue #412, opened 2021-11-03
Per component focus border


ebourg:
The focus border configured with `Component.focusedBorderColor` is applied to all components. It would be nice to be able to disable it for some components such as the JScrollPanes.

DevCharly:
You can disable focus border for single scroll pane:

~~~java
Component view = myScrollPane.getViewport().getView();
((JComponent)view).putClientProperty(
    FlatClientProperties.COMPONENT_FOCUS_OWNER,
    (Predicate<JComponent>) c -> false );
~~~

ebourg:
Thank you for the tip, I guess making the scroll pane non focusable would work too, but I was looking for something more like a `ScrollPane.focusedBorderColor` property overriding the default `Component.focusedBorderColor`.

DevCharly:
There is actually a simple solution.

Create a own border class that changes the color:
~~~java
public class MyScrollPaneBorder extends FlatBorder {
    public MyScrollPaneBorder() {
        focusedBorderColor = Color.red;
    }
}
~~~

And install it:
~~~java
MyScrollPaneBorder myScrollPaneBorder = new MyScrollPaneBorder();
UIManager.put( "ScrollPane.border", myScrollPaneBorder );
UIManager.put( "Table.scrollPaneBorder", myScrollPaneBorder );
~~~

PR #417, opened 2021-11-05
Styling: support component type specific styles

DevCharly:
PRs #341 and #388 added basic styling and style classes E.g.
~~~java
mySlider.putClientProperty( "FlatLaf.style", "thumbColor: #f00" );
or
mySlider.putClientProperty( "FlatLaf.styleClass", "myclass" );
~~~
which is comparable to what can be done in HTML:
~~~html
<p style="background: #f00">
or
<p class="myclass">
~~~

This PR is the next step and adds support for **component type specific styles**, 
which are **applied to all components of that type**.

E.g. in FlatLaf properties file:
~~~
[style]ScrollPane = focusedBorderColor: #f00
~~~
or in Java code:
~~~java
UIManager.put( "[style]ScrollPane", "focusedBorderColor: #f00" );
~~~

This style is applied to all scrollpanes in the application.

This is comparable to what can be done in CSS. E.g.
~~~css
p {
    background: #f00;
}
~~~

## Issues/problems

Because those style rules are applied to all components of a type within the application, this may cause some side effects or unwanted changes.

E.g. some components internally use other components (`JButton` for arrow button in `JComboBox`; `JTextField` in editable `JComboBox`) and those internal components are changed too.

Some styling properties are applied to borders (e.g. see fields in class `FlatBorder` annotated with `@Styleable`). When the border of a component is removed/changed then those style properties no longer can applied and exceptions are logged.

~~~java
UIManager.put( "[style]ScrollPane", "focusedBorderColor: #f00" );

JScrollPane scrollPane = new JScrollPane();
scrollPane.setBorder( BorderFactory.createEmptyBorder() );
~~~

The above code initially works because the style rule is applied within the constructor and the border is changed after it, but if you switch the theme, the style rule `focusedBorderColor: #f00` can not applied to the `EmptyBorder` and an `UnknownStyleException` is logged.

FlatLaf is very strict (at the moment) and logs any style property that can not applied.
CSS on the other side simply ignores unknown styles...

CC @ebourg  (issue #412); @Chrriis (issue #117); @orange451 (issue #340)

I'm a little bit undecided whether this PR is a good idea, or do more harm than good?

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatStylingSupport.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatStylingSupport.java
@@ -485,22 +485,34 @@ public class FlatStylingSupport
 		String getterName = buildMethodName( "get", name );
 		String setterName = buildMethodName( "set", name );
 
-		try {
-			Method getter;
+		for(;;) {
 			try {
-				getter = cls.getMethod( getterName );
+				Method getter;
+				try {
+					getter = cls.getMethod( getterName );
+				} catch( NoSuchMethodException ex ) {
+					getter = cls.getMethod( buildMethodName( "is", name ) );
+				}
+				Method setter = cls.getMethod( setterName, getter.getReturnType() );
+				Object oldValue = getter.invoke( obj );
+				setter.invoke( obj, convertToEnum( value, getter.getReturnType() ) );
+				return oldValue;
 			} catch( NoSuchMethodException ex ) {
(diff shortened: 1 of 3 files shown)


Chrriis:
My original issue was to use a FlatLaf theme but to have a custom component somewhere.
I am not sure which use case the feature you are presenting is trying to solve. Is it a way to have in effect multiple FlatLaf themes? Or is it a way to alter a theme after it is loaded, but then why not alter the theme?

PR #429, opened 2021-11-19
Window title bar improvements (Windows 10/11 only)

DevCharly:
This PR brings some improvements to the window title bar used in FlatLaf window decorations.
This affects only Windows 10/11.

## Unified background

Because the trend goes to "gray" window title bars on Windows 11, the FlatLaf "unified background" for window title bar is now enabled by default. The window title bar has now same background color as window content. Bottom separator for menu bars is no longer painted (if unified background is enabled).

image

If you prefer the old behavior use:

~~~java
UIManager.put( "TitlePane.unifiedBackground", false );
~~~

## Icon in window title bar

It is now possible to hide window title bar icon.

image

For single window use:

~~~java
myFrame.getRootPane().putClientProperty( "JRootPane.titleBarShowIcon", false );
~~~

For all windows in the app use:

~~~java
UIManager.put( "TitlePane.showIcon", false );
~~~

**Note**: It is always a good idea to apply an icon to the main frame (via `JFrame.setIconImages(List<? extends Image> icons)`) because this icon is also displayed in Windows taskbar and when switching applications via `Alt+Tab`.

## JOptionPane title bar icon

Previously, the icon of the parent window was shown in `JOptionPane` dialogs title bar. This was different to "standard" Swing, which did not show an icon. With this PR, FlatLaf also no longer shows an icon in `JOptionPane` dialogs title bar.

image

If you prefer the old behavior use:

~~~java
UIManager.put( "OptionPane.showIcon", true );
~~~


eirikbakke:
Random idea: As an alternative to hiding the icon entirely, it might be nice to have the option to display the icon next to the centered title, e.g. like this (mockup):

iconmiddle

DevCharly:
@eirikbakke thanks for the idea. It is now implemented in PR #565

PR #440, opened 2021-12-07
Added Jailer to Applications using FlatLaf

Wisser:
Hi,

I would like to add my tool [Jailer ](https://github.com/Wisser/FlatLaf) to the list.


DevCharly:
Thanks 👍

PR #449, opened 2021-12-19
Restored default constructors for UIs

ingokegel:
The introduction of the UI constructors with a `shared` boolean flag has removed the implicit default constructors. That change broke backwards compatibility for derived classes.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java
@@ -158,6 +158,10 @@ public class FlatButtonUI
 			: new FlatButtonUI( false );
 	}
 
+	public FlatButtonUI() {
+		this( false );
+	}
+
 	/** @since 2 */
 	protected FlatButtonUI( boolean shared ) {
 		this.shared = shared;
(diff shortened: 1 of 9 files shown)


DevCharly:
This was actually an intentional incompatibility to ensure that derived classes are adopted.

FlatLaf does not know whether derived UI delegate classes are shared or not.
For styling (see PR #341 and #388), a mechanism to "unshare" a shared UI delegate is needed.
This is done with some modifications to method `createUI()`. E.g.:

https://github.com/JFormDesigner/FlatLaf/blob/840083940da46dd8baa1513c2613641b5e183322/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java#L155-L164

Sure, it is possible that derived classes use a simple `createUI()` method, like following, and sharing is no issue.
~~~java
 public static ComponentUI createUI( JComponent c ) { 
 	return new MySpecialFlatButtonUI();
 }
~~~

Besides that it is necessary to update source code of derived UI delegate classes,
is there another issue with the changed constructor?


### Regarding API compatibility

Only classes in packages `com.formdev.flatlaf` and `com.formdev.flatlaf.util` are kept API compatible.
There is a sigtest task in `flatlaf-core/build.gradle.kts` that automatically checks it.
(the main motivation for this is a NetBeans core plugin).

Due to future development, "API" of classes in all other package may change (without notice).

While it is fine to extend UI delegate classes (e.g. `FlatButtonUI`) or use classes like `FlatUIUtils`, I can not guarantee API compatibility for those. Adding new features sometimes makes it necessary to change something (e.g. add a parameter to a method). But I avoid unnecessary changes like renaming classes/methods/fields.

ingokegel:
Thank you for the detailed explanation. I mistakenly thought that the change was unintentional.

PR #453, opened 2021-12-21
Add a system property to load pre-extracted native libraries from a directory

ingokegel:
Added the system property flatlaf.nativeLibraryPath to specify a directory where the native libraries have been extracted.

Avoiding extraction at runtime to the temporary directory is useful in order to prevent anti-virus software from blocking the library loading.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatSystemProperties.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatSystemProperties.java
@@ -140,6 +140,12 @@ public interface FlatSystemProperties
 	String USE_TEXT_Y_CORRECTION = "flatlaf.useTextYCorrection";
 
 	/**
+	 * Specifies a directory in which the native FlatLaf library have been extracted.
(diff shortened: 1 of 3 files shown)


ingokegel:
Btw, we still use https://github.com/JFormDesigner/FlatLaf/pull/278 in install4j for the generated installer itself where we shade FlatLaf into a different package (so the default DLL would not work anyway) and where we do not want any DLLs to be extracted into the temporary directory.

The feature in this pull request is for the applications themselves where we use the standard FlatLaf distribution. We used JetBrains window decorations so far, but to get support for the Windows snap layout, we have to turn them off and so the extraction of the library to the temporary directory became an issue.

DevCharly:
Many thanks. Merged.

In commit 221e801561c92e9ed7e0411dd24bc1db8b3a1d41, I've replaced `libraryFile.getPath()` with `libraryFile.getAbsolutePath()`.
This allows using a relative path in `flatlaf.nativeLibraryPath`.
E.g. `-Dflatlaf.nativeLibraryPath=lib/natives`.

PR #484, opened 2022-02-19
Fix #483: NPE when painting icon on OS X top menu bar

### LINKED ISSUE #483: NPE when painting icon on OS X top menu bar
Painting on the native menu bar triggers a NPE because [the icon is painted with a null component](https://github.com/openjdk/jdk/blob/d7f31d0d53bfec627edc83ceb75fc6202891e186/src/java.desktop/macosx/classes/com/apple/laf/AquaIcon.java#L68).

nroduit:
See #483


DevCharly:
Thanks for your PR.

I've merged it, but without the "float operation" fixes.
(that the reason that GitHub does not mark it as merged)

The float operation fixes are not necessary/wanted because width/height are even numbers.
And in case of changing them to uneven numbers, it is be better to paint the maximize rectangle at whole pixels.

PR #522, opened 2022-04-27
File chooser shortcuts panel

DevCharly:
This PR adds a shortcuts panel to the left side of all file choosers (class `JFileChooser`):

image

image

On Windows, the content of the shortcuts panel is the same as in the Java Windows L&F.
On macOS and Linux the shortcuts panel is empty and hidden.

However you can customize the content. E.g.

~~~java
UIManager.put( "FileChooser.shortcuts.filesFunction", (Function<File[], File[]>) files -> {
    ArrayList<File> list = new ArrayList<>( Arrays.asList( files ) );
    list.add( 0, new File( System.getProperty( "user.home" ) ) );
    return list.toArray( new File[list.size()] );
} );

UIManager.put( "FileChooser.shortcuts.displayNameFunction", (Function<File, String>) file -> {
    if( file.getAbsolutePath().equals( System.getProperty( "user.home" ) ) )
        return "Home";
    return null;
} );
UIManager.put( "FileChooser.shortcuts.iconFunction", (Function<File, Icon>) file -> {
    if( file.getAbsolutePath().equals( System.getProperty( "user.home" ) ) )
        return new FlatFileChooserHomeFolderIcon();
    return null;
} );
~~~

image

### Downsides

The get the list of shortcut folders and to get the large icons, FlatLaf uses API that is not available in all Java versions.

- the Java API to get list of shortcut folders is available since Java 12; reflection is used for older Java versions
- the Java API to get large icons is available since Java 17; reflection is used for older Java versions

So for Java 9 to 16 you'll get following warning on stdout when a file chooser is opened.

~~~
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.formdev.flatlaf.ui.FlatFileChooserUI$FlatShortcutsPanel (file:...) to method sun.awt.shell.ShellFolder.get(java.lang.String)
WARNING: Please consider reporting this to the maintainers of com.formdev.flatlaf.ui.FlatFileChooserUI$FlatShortcutsPanel
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
~~~

If you use `--illegal-access=deny`, then the shortcuts panel will be empty (Java 9 to 11),
and small icons are scaled up (Java 9 to 18).

CC @Chrriis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 FlatLaf Change Log
 ==================
 
+## 2.3-SNAPSHOT
+
+#### New features and improvements
+
+- FileChooser: Added (optional) shortcuts panel. On Windows it contains "Recent
+  Items", "Desktop", "Documents", "This PC" and "Network". On macOS and Linux it
+  is empty/hidden. (issue #100)
+
+
 ## 2.2
 
 #### New features and improvements
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatFileChooserUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatFileChooserUI.java
@@ -22,6 +22,7 @@ import java.awt.Dimension;
 import java.awt.Graphics;
 import java.awt.Graphics2D;
 import java.awt.Insets;
+import java.awt.LayoutManager;
 import java.awt.RenderingHints;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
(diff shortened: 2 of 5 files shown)


Chrriis:
There are several expectations with file choosers:
1. Faster navigation with shortcuts and perhaps some who want to customize these shortcuts.
2. People who want a more native file chooser in line with the look and feel.
3. People who want the native file chooser, independent from the look and feel.

I don't think FlatLaf should care about item 3 because there are other libraries that do it and it breaks the idea of a look and feel (I saw discussions on that topic lately, this is just my opinion).

This PR is more about 1 and 2:
- It does give people a way to customize the file chooser shortcuts.
- It feels more native because it has a navigation pane.

That being said, and correct me if I am wrong, it looks like the Windows 7 file chooser, or the AWT file chooser. All the file choosers that I can see on my Windows 10 have a tree on the left. My native file chooser are more like the second one from this (random) link:
https://stackoverflow.com/questions/39819319/windows-native-file-chooser-in-java

So, it is definitely an improvement, it solves item 1, but it is a partial answer to item 2. I understand the constraints so I know that it is not possible to support item 2 completely (like having a tree on the left, and editable crumb bar at the top).

I do welcome this change! Keep up the good work!

PR #565, opened 2022-07-11
Window title bar usability improvements (Windows 10/11 only)

DevCharly:
This PR brings some usability improvements to the window title bar used in FlatLaf window decorations.
This affects only Windows 10/11.

## Window resizing

There are now additional areas at top of embedded menu bar to resize the window. This makes it easier to resize small frames with lot of menus.

image

## Minimum window title width for moving window

The window title now has a minimum width, which avoids that it is completely hidden in small frames with menus. The window title is important because it is needed to move the window. Without any window title, you can move the window only via the small area around the window icon.

image

## Buttons shrink if space is rare

The width of iconify/maximize/close buttons is reduced if space is rare, to give more space to embedded menu bar and title.

image

image

## Other changes

- no longer reduce height of window title bar if it has an embedded menu bar and is maximized
- using larger gaps and minimum sizes
- added minimum gap between embedded menu bar and window title
- fixed oscillating title while resizing window width
- fixed lost right-to-left component orientation in title bar when switching Laf


DevCharly:
## Show app icon beside title

There is a new option to show window icon beside window title, if menu bar is embedded or title is centered.

image

Enable with:

~~~java
UIManager.put( "TitlePane.showIconBesideTitle", true );
~~~


If window title centering is enabled (`TitlePane.centerTitle = true`) then the icon is also centered.

image

PR #579, opened 2022-08-21
Linux window decorations: native move window and system menu

DevCharly:
This PR improves the behavior of FlatLaf custom window decorations on Linux to bring it closer to system window decoration behavior:

- when moving/dragging a window, now Linux window manager (WM) is used for the drag operation. This enables features provided by Linux WMs. E.g. maximize window by dragging it to the top of the screen. Or dragging window to one side of screen to maximize it along the side. Or snap to other windows while dragging.
- right-click on window title bar now shows the window system menu (only on GNOME; not available on KDE)

image

Tested on CentOS 7, Fedora 35, Ubuntu 20.04, Ubuntu 22.04 and Kubuntu 22.04 (KDE).
Tested with Java 8, 11, 17 and 18 (openjdk.java.net and Adoptium).

To try it out, download the artifacts here:
https://github.com/JFormDesigner/FlatLaf/suites/7910208039/artifacts/336738762


DevCharly:
On KDE, window system menu (on right-click) probably works starting with **KDE 5.96.0** (released 9 July 2022),
which implements `_GTK_SHOW_WINDOW_MENU`: https://kde.org/announcements/frameworks/5/5.96.0/#kwindowsystem

matthiasblaesing:
Nice - implementation looks straight forward and a quick test was successful.

PR #595, opened 2022-09-25
Switch and close tabs on left mouse click only

ShadelessFox:
It feels more natural when right/middle mouse clicks don't do anything when you click on a tab.


webfolderio:
This should be configurable.

DevCharly:
> This should be configurable.

Why?

DevCharly:
> It feels more natural when right/middle mouse clicks don't do anything when you click on a tab.

Thanks. You're right.
Nearly all Swing components respond only to left clicks.
Only exception I've found so far are split pane divider and menus.

I think it is a bug in all Swing L&F to switch tabs on right/middle mouse click.
This becomes a real problem when you add a mouse listener to the tabbed pane and
want show a context menu for tabs on right-click.
Without this PR, the right-clicked tab is always selected. There was no way to prevent this.
With this PR, the app can control whether it want select the clicked tab or not.

Tried several Windows and macOS apps that have tabs.
All show a context menu on clicked tab, but do not select/activate that tab.

PR #605, opened 2022-10-29
FlatInspector: add/remove mouse listener in setEnabled

abextm:
Currently there isn't a good way to get rid of the mouse listener the inspector adds. Alternatively the mouse listener could be added/removed in `setEnabled`, in which case `uninstall` could go away entirely.

### DIFF
--- a/flatlaf-extras/src/main/java/com/formdev/flatlaf/extras/FlatInspector.java
+++ b/flatlaf-extras/src/main/java/com/formdev/flatlaf/extras/FlatInspector.java
@@ -155,7 +155,7 @@ public class FlatInspector
 						rootPane.putClientProperty( FlatInspector.class, inspector );
 						inspector.setEnabled( true );
 					} else {
-						inspector.uninstall();
+						inspector.setEnabled( false );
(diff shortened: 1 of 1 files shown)


DevCharly:
Thanks for pointing out this issue.

Seems to occur only if you use something like:
~~~java
FlatInspector inspector = new FlatInspector( rootPane );
inspector.setEnabled( true );
...
inspector.setEnabled( false );
~~~

When using following and enable/disable inspector via shortcut keys, `uninstall()` is invoked and the mouse listener removed.
~~~java
FlatInspector.install( "ctrl shift alt X" );
~~~

> Alternatively the mouse listener could be added/removed in `setEnabled`, in which case `uninstall` could go away entirely.

That would be much better 👍 
The other listeners are also added/removed in `setEnabled()`.
Don't know why mouse motion listener is added in constructor and removed in `uninstall()`.
Maybe a refactoring error...

Would you like to update the PR?

abextm:
Done

DevCharly:
Perfect 👍
Thx

PR #669, opened 2023-04-20
Adding support for the Palette of Netbeans

SoftwareOrgMx:
Adding support for the Palette of Netbeans. Now when you add FlatButton you will see a button Icon.

Screenshot

This pull is a example if you accept it i will continue with other components.


DevCharly:
Sorry, but don't want to have such BeanInfos in FlatLaf for various reasons.

One reason is that I'd like to encourage you to improve NetBeans instead and
implement the same method that [JFormDesigner](https://www.formdev.com/jformdesigner/) uses to get icons:
if `BeanInfo.getIcon(int type)` returns `null`, then get icon from BeanInfo(s) of component superclass(es).
Then `FlatButton` would automatically use icon from `JButton`.

Here is the where the NetBeans palette gets the icon:
https://github.com/apache/netbeans/blob/e65310a45ade2f08ce756536ce83bcc49237b101/java/form/src/org/netbeans/modules/form/palette/PaletteItem.java#L271

For the icons in the NetBeans Navigator tree, the code is here:
https://github.com/apache/netbeans/blob/e65310a45ade2f08ce756536ce83bcc49237b101/java/form/src/org/netbeans/modules/form/RADComponentNode.java#L132

PR #671, opened 2023-04-22
Add Kafka Visualizer to Applications

kumait:
Kafka Visualizer is a Kafka GUI client that uses FlatLaf, I added it to the list of applications using FlatLaf.


DevCharly:
Thanks

Issue #664, opened 2023-04-10
Possibly better support for SVG icons


weisJ:
As I understand it currently SVG icons are rendered using the svgSalamander library. Sadly it isn't under active development anymore and won't return to it anytime soon. I have been developing an alternative to it (https://github.com/weisJ/jsvg) and in the meantime it has become almost on par feature-wise with svgSalamander in some areas and exceeds it in some others. In particular the only missing features I am aware of are animations and element picking by position which aren't exposed through `FlatSVGIcon` anyway.

Note: The Jetbrains team recently switched from rendering their UI icons with Batik to JSVG, so the library seems to work well for them:
https://github.com/JetBrains/intellij-community/commit/93a8fe3d8376a1ee4b4bcf4cf85d991afc01013d

I was wondering whether replacing svgSalamander with JSVG would be something to consider here? If yes I would be happy to prepare a PR :)

weisJ:
I have released version 1.0.0 and submitted a PR. There were some outstanding performance and correctness issues with commonly used filters which I wanted to get fixed first hence the wait.

DevCharly:
JSVG is coming to FlatLaf 3.2 😄

PR #684, opened 2023-05-29
Replace svgSalamander with jsvg

weisJ:
See #664

### DIFF
--- a/flatlaf-extras/README.md
+++ b/flatlaf-extras/README.md
@@ -4,8 +4,7 @@ FlatLaf Extras
 This sub-project provides some additional components and classes:
 
 - [FlatSVGIcon](https://www.javadoc.io/doc/com.formdev/flatlaf-extras/latest/com/formdev/flatlaf/extras/FlatSVGIcon.html):
-  An icon that displays SVG using
-  [svgSalamander](https://github.com/JFormDesigner/svgSalamander).\
+  An icon that displays SVG using [JSVG](https://github.com/weisJ/jsvg).\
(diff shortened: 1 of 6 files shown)


rogerbj:
@weisJ Great work! Just tried the **FlatLafDemo** and if any `ColorFilter` has been set using `FlatSVGIcon.setColorFilter()` or `FlatSVGIcon.ColorFilter.setMapper()`, nothing happens with jsvg. This can be tested in the **FlatLafDemo** and in the **Extras** tab.

rogerbj:
@weisJ I also ran into a problem in `FlatSVGIcon.loadFromStream()`:

```URI dummyUri = new URI( "inputStreamSVG", "/flatlaf-stream-" + streamNumber++, null );```

It produces an **"Invalid protocol: inputStreamSVG"** error. 

Except for this and my other report it works a lot better with randomly tested SVGs as opposed with SvgSalmander that fail with those.

DevCharly:
@weisJ many thanks for the PR and for JSVG 👍 
@rogerbj thanks for testing.

I've reviewed it and fixed in color filter and loading from input stream issues.
Have also simplified the code a little bit by replacing internal use of URI with URL.
(URI was used/required for svgSalamander)

Will test this PR and JSVG in some other projects in the next weeks...

weisJ:
Note that I had deliberately not changed the cache to take URLs as keys due to this problem: https://errorprone.info/bugpattern/URLEqualsHashCode

DevCharly:
Many thanks for this hint. Didn't know that. Changed the cache key from URL to String to avoid that.

PR #695, opened 2023-06-26
Issue 689 system.get property security exception

rcuprak:
A proposed fix for Issue 689. I have added a class FlatSystemCachedProperties which caches values returned by calls to System.getProperty(). Use of the cache can be turned on/off. So if you are going to be executing code with a security manager, you can enable it and cached values will be used.


DevCharly:
This is a bug in OpenWebStart: https://github.com/karakun/OpenWebStart/issues/76

Issue #403, opened 2021-10-26
Invoking JFileChooser with FlatLaF crashes VM


WLenhard:
With the latests Adoptium Release (jdk-17+35), installing FlatLaF and invoking a JFileChooser leads to fatal crash of the VM. It did not happen with jdk-16 and I am not sure, if this is a problem with FlatLaF or Adoptium. OS: Win10 Enterprise 19042.1288

Minimal reproducible example:

```

import javax.swing.JFileChooser;
import javax.swing.SwingUtilities;
import com.formdev.flatlaf.FlatLightLaf;

public class FileChooserTest {
public static void main(String[] args) {
	
	FlatLightLaf.setup();
	
	SwingUtilities.invokeLater(new Runnable() {
		@Override
		public void run() {
			try {
				JFileChooser chooser = new JFileChooser();
				chooser.showSaveDialog(null);
				String file = chooser.getSelectedFile().getName();
				System.out.println(file);
			} catch (Exception e) {
				e.printStackTrace();
			}

		}
	});
}
}

```

WLenhard:
Hi Charly,
I checked with the new JRE17.0.1+12 and can now trace back the error a little bit better. The error still occurs, but only in the x86 runtime and not in x64 on Windows. The runtime crashes when opening a JFileChooser. Other Swing components seem to work fine.

Best regards!

DevCharly:
Can reproduce it.
It is not related to FlatLaf.
Occurs also in Windows L&F.
Metal L&F works because it does not show system icons.

It is a bug in OpenJDK when accessing system icons for the file chooser.
Occurs also in Windows L&F. Metal L&F does not show system icons.

There was a change in this area in JDK 17 to access "large Windows icons", which may be responsible for this crash.
See: https://bugs.openjdk.java.net/browse/JDK-8182043

Maybe this change was never tested on 32 Bit because Oracle distributes only 64 Bit JDKs?

DevCharly:
To avoid the crash I've implemented an workaround, which simply does not use Windows system icons on 32bit Java 17...

64-bit Java 17 is not affected.

DevCharly:
Workaround is in [1.6.2](https://github.com/JFormDesigner/FlatLaf/releases/tag/1.6.2).

DevCharly:
related JDK bug: https://bugs.openjdk.java.net/browse/JDK-8277299

Fixed in Java 18 EA 32bit

[...]

PR #709, opened 2023-08-06
x86: Narrow version range for not using system icons

syoon2:
Related: #403

The fix for [JDK-8277299](https://bugs.openjdk.org/browse/JDK-8277299) was backported to Java 17.0.3 (with https://github.com/openjdk/jdk17u-dev/pull/98) and using system icons in later subversions should not cause a crash. Thus, this PR amends the version range for disabling system icons.


DevCharly:
Many thanks for this PR.

While testing, I found out that FlatLaf application crash with 32-bit Java 17 - 17.0.2 on Windows caused by PR #522. This is fixed in commit 4b4990635d2286a23b6b98c029821c0a149efdee

PR #713, opened 2023-08-13
Scrollpane rounded border

DevCharly:
This is an attempt to bring rounded border to `JScrollPane` and all component that are usually placed into a scrollpane. E.g. `JTextArea`, `JTextPane`, `JEditorPane`, `JList`, `JTree` and `JTable`.

The problem is that there are multiple overlapping components involved:
- the `JScrollPane,` which has the border
- the `JViewport,` which has the view component as child (e.g. a `JList`)
- optional column and row headers
- optional corner components
- vertical and horizontal `JScrollBar`s

They are painted in this order and using rounded border on scrollpane would look like this:

grafik

The viewport/list and the scrollbar simply paint over the rounded corners.
And there is no way to fix this for the viewport in the look and feel.
(maybe a custom scrollpane implementation, that paints its border and corners after its children, could work)

So the idea to solve this problem is to add some extra space at left and right sides between the border and the viewport/scrollbars:

grafik

That's better, but the scrollbar should be at the border:

grafik

We can hide the scrollbar track and make the thumb round to improve the look:
~~~properties
ScrollBar.track = @componentBackground
ScrollBar.thumbInsets = 2,2,2,2
ScrollBar.thumbArc = 999
~~~

grafik

If the vertical scrollbar is hidden, it looks like this:

grafik

With rounded selection (and smaller border arc than in previous screenshots) this looks really nice:

~~~properties
ScrollPane.arc = 12
List.selectionArc = 8
~~~

grafik grafik

Here are some screenshot that shows the extra space, which was added for rounded border, in green:

grafik grafik

Smaller border arc results in smaller extra space, larger arc gives larger extra space:

grafik grafik

Possible improvements/ideas:

- ~~make viewport wider if vertical scrollbar is visible to remove the gap between viewport and vertical scrollbar (see above screenshot)~~
- adapt vertical scrollbar painting to rounded border to allow full height vertical scrollbar
- ~~allow specifying scrollpane border arc for component types (e.g. use rounded border for multi-line text components, but not for lists, tables and trees)~~

For testing you need to enable rounded border. E.g.:

~~~properties
ScrollPane.arc = 12
~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 FlatLaf Change Log
 ==================
 
+## 3.3-SNAPSHOT
+
+#### Fixed bugs
+
+- ScrollPane: Styling ScrollPane border properties did not work if view
+  component is a Table.
+
+
 ## 3.2
 
 #### New features and improvements
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatBorder.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatBorder.java
@@ -28,7 +28,6 @@ import javax.swing.JComboBox;
 import javax.swing.JComponent;
 import javax.swing.JScrollPane;
 import javax.swing.JSpinner;
-import javax.swing.JViewport;
 import javax.swing.UIManager;
 import javax.swing.plaf.basic.BasicBorders;
 import com.formdev.flatlaf.FlatClientProperties;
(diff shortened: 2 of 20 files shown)


DevCharly:
Last commit removes gap between viewport and vertical scrollbar (if possible).

Before:

grafik

After:

grafik

If the scrollpane has a column header (e.g. a JTableHeader when using JTable) and a upper-right corner component (red rectangle), then it is not possible to remove the gap:

grafik

BrantLauro:
Very good work, Can I already use that as a Snapshot or something like this?

DevCharly:
Last commit adds possibility to specify scroll pane arc for multi-line text components, lists, tables and trees.

There are four new UI properties that have default value `-1`, which means that `ScrollPane.arc` is used.

~~~properties
ScrollPane.List.arc = -1
ScrollPane.Table.arc = -1
ScrollPane.TextComponent.arc = -1
ScrollPane.Tree.arc = -1
~~~

E.g. if you want rounded border only for multi-line text components, then simply use:

~~~properties
ScrollPane.arc = 0
ScrollPane.TextComponent.arc = 12
~~~

Issue #715, opened 2023-08-16
macOS: Rounded popup windows


ctipper:
I run a cross-platform app that uses Flatlaf for macOS, Windows and Linux. On macOS I am using FlatMacLightLaf and FlatMacDarkLaf on other platforms I am using FlatLightLaf and FlatDarkLaf with theme switching. PR #643 introduced rounded popup windows but only for FlatLightLaf etc. I'd like this made the default in FlatMacLightLaf etc. This would align with the rounded menus and selection highlights in the macOS themes.

In my code the following sets the properties but there is no effect when using FlatMacLightLaf etc.

```
/**
 * @see <a href="https://github.com/JFormDesigner/FlatLaf/pull/643">Windows 11: Rounded popup windows #643</a>
 */
UIManager.put("PopupMenu.borderCornerRadius", 8);
UIManager.put("ComboBox.borderCornerRadius", 8);
UIManager.put("ToolTip.borderCornerRadius", 8);
UIManager.put("Popup.borderCornerRadius", 8);
```

ctipper:
I know it looks like an error, but I can say that these properties have long been ignored and I don't think the macos theme has ever had rounded combo boxes and tooltips in my dev environment. Now on Apple Silicon maOS 13.5.1 , but previously Intel macOS 12.6

ctipper:
No rounded drop downs on Linux either, Debian 12 JDK 17

DevCharly:
PR #643 works only on **Windows 11**. See title and description of this PR.
Not sure why you expect that this also works on macOS or Linux 😕 

Java/Swing does not support native rounded borders out-of-the-box.
PR #643 uses a native library (DLL) to modify the Win32 window, that Java created, to enable rounded border.

To do similar on macOS, a new native library for macOS is required that does modify the macOS windows.
I already know that this is possible and roughly how to do this, but implementing and testing this is time consuming.
I expect at least 3-4 days to make this work.

If you're working for a company, please ask you boss whether the company is willing to sponsor FlatLaf to finance development of this feature. Estimated cost: 2000 EUR/USD. https://www.formdev.com/flatlaf/sponsor/
I could work on it in September and release it in FlatLaf 3.3.

Sorry to ask for sponsorship, but can't implement all the features companies want for free...

DevCharly:
Let's keep the issue open. Maybe some company passes by and is willing to sponsor development of this feature.

DevCharly:
See PR #772

PR #772, opened 2023-12-09
macOS: native rounded borders for popups

DevCharly:
This PR enables **native macOS rounded borders** (including drop shadow) for popups. (issue #715)

PopupMenu:

image

ComboBox (macOS theme):

image

ToolTip:

image

The rounded popup borders are enabled by default for all themes and
use the same UI properties and client property as in PR #643.

Additionally there are 4 new UI properties:
~~~
PopupMenu.roundedBorderWidth
ComboBox.roundedBorderWidth
ToolTip.roundedBorderWidth
Popup.roundedBorderWidth
~~~

The first three are used for `JPopupMenu`, `JComboBox` and `JToolTip`.
The last one is used for other components shown as popup via `PopupFactory.getPopup()`.

There is also a client property `Popup.roundedBorderWidth` that allows changing rounded border width for specific popups.

`roundedBorderWidth` is used only on macOS.
The default value is `0` for light themes (because macOS always adds a thin dark border),
and `1` for dark themes (to add a lighter border to the popup, otherwise it is difficult to distinguish the popup from the background).


DevCharly:
I've tested this only on macOS 10.14 (Intel) and on 13.3 (M1).

Would be great to get some feedback whether this works without problems on other macOS versions.

Prebuilt binaries are available here (requires GitHub login):
https://github.com/JFormDesigner/FlatLaf/suites/18901204850/artifacts/1103682403

PR #790, opened 2024-01-09
Table header background is never restored after hover when the renderer background is null

dar-dev:
The table header background is never restored after hover when the label renderer is opaque. 

Here's a demo video:

https://github.com/JFormDesigner/FlatLaf/assets/55930104/3c154be6-f2b3-439c-bab9-c893eb865505

Cheers,

Dar

--

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTableHeaderUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTableHeaderUI.java
@@ -360,6 +360,9 @@ public class FlatTableHeaderUI
 			Icon oldIcon = null;
 			int oldHorizontalTextPosition = -1;
 
+			// needed when we have to restore the label background after hover and the renderer is opaque
+			boolean restoreBackground = false;
+			
 			// hover and pressed background/foreground
 			TableColumn draggedColumn = header.getDraggedColumn();
 			Color background = null;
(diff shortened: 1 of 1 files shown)


DevCharly:
> ... when the label renderer is opaque.

Sorry, but I'm not sure what you mean. 
Is the renderer's background color `null`?

Could you please extend following test case to demonstrate the issue? Thanks.

~~~java
import javax.swing.*;
import javax.swing.table.DefaultTableModel;
import com.formdev.flatlaf.FlatLightLaf;

public class TableHeaderTest
{
	public static void main( String[] args ) {
		SwingUtilities.invokeLater( () -> {
			FlatLightLaf.setup();

			JTable table = new JTable( new DefaultTableModel( 3, 3 ) );
			JScrollPane scrollPane = new JScrollPane( table );

			JFrame frame = new JFrame( "Test" );
			frame.setDefaultCloseOperation( WindowConstants.DISPOSE_ON_CLOSE );
			frame.add( scrollPane );

			frame.setSize( 400, 300 );
			frame.setVisible( true );
		} );
	}
}
~~~

DevCharly:
Thanks, now I understand 😄 

Background and foreground colors are set to `null` in `DefaultTableCellRenderer.updateUI()`.
`DefaultTableCellRenderer.getTableCellRendererComponent()` would change those colors to non-null values,
but this method is overridden. So they stay `null` and FlatLaf does not restore them...

PR #794, opened 2024-01-13
FlatMenuUI: remove JMenuBar dependency

rogerbj:
A **ClassCastException** is thrown in `FlatMenuUI.java` when using a `CommandMenuBar.java` in JIDE. This occurs since `FlatMenuUI` assumes there is always a `JMenuBar` as parent for a `JMenuItem`. The problem occurs if resizing the width of the window whereas JIDE automatically replaces `JMenu`'s that doesn't fit with a clickable icon. Clicking it shows a popup window with the menus that doesn't fit to display. The owner of the `JMenuItem`'s is then a `CommandBarPane` instance rather than the original `CommandMenuBar`. The CCE happens when clicking the icon.

image

This PR generalizes `FlatMenuUI` to not require a `JMenuBar`.

With this change it now works fine with JIDE.


DevCharly:
Thanks for the PR.

I wonder in which situation the `ClassCastException` is thrown?

`CommandMenuBar` extends `JMenuBar`.
So as long as the `JideMenu` is a direct child of the `CommandMenuBar`, it should work fine.

rogerbj:
@DevCharly I've updated the original PR description with a sample screenshot that uses this PR branch.

DevCharly:
Thx. Finally I understand why `JideMenu` overrides `isTopLevelMenu()` 😄 

Instead of merging this PR, I've cherry picked your commit 4d4b90c989f2de9f5474bcd13281c6841b513749.
It is now in FlatLaf/main in commit 0c00117820ff65bf814c7f8db0c2d97087782486.

This PR contains a lot of "Merge remote-tracking branch 'origin/master'" commits,
which I did not like to add to the history of this repo.

For future PRs, please base them on main branch of this repo.

To update dbvis/FlatLaf repo, it would be better to avoid merging from this repo.
Better fetch commits from this repo to a local FlatLaf repo, do a local fast-forward merge and then push to dbvis/FlatLaf repo.

rogerbj:
@DevCharly Super! Yeah, sorry the PR did include a lot of commits. Will be more careful in the future.

PR #817, opened 2024-03-10
Add LinearGradient paint in svg color filter

DJ-Raven:
When svg element paint the linearGradient, The filter color not work, also gray filter use when compnent is disable, This PR I have fixed this can you please review it, Thank you

JButton Enable

JButton Disable


DevCharly:
Thank you 👍

Issue #648, opened 2023-02-09
FlatLaf and Proguard ? 


x9ware-llc:
Has anyone used Proguard against FlatLaf? I currently just keep everything since I was unable to successfully apply any optimization or obfuscation. Here is the keep everything that in my existing definition:  -keep class com.formdev.** { *; }. I can obviously drop the line number table and then I also use -injars '....[filereference]....\flatlaf-2.6.jar'(!com/formdev/flatlaf/natives/**) to drop the embedded DLLs. However, it would be great to be able to do more if possible.

x9ware-llc:
I hear you and agree on all points. We reduced the FlatLaf jar size by about 140kb when we dropped the DLL and line numbers. I asked the question really just to see if anyone else is using ProGuard with FlatLaf. We use it against our entire application fatjar, with some overrides needed for things like Apache POI and JAXB. Completely excluding FlatLaf from the obfuscation makes it a bit of an outlier; it would be nice if it could be included. I realize that there is no intellectual property requirement, and that there is probably no more than a 100kb reduction to be obtained. I also concur that a first step would be to enable optimization with obfuscation turned off.

Andy-DG:
How would I make sure I exclude flatlaf from obfuscation?
I tried with:
-keep class com.formdev.flatlaf.** { *; }
But that did not seem to resolve the problem.

x9ware-llc:
I am using these Proguard statements. There is not much to be gained here, since obviously FlatLaf is small in the first place. I am also preprocessing the FlatLaf jar to drop line numbers and the DLL, since I would rather have the DLL excluded. Dropping the DLL works with 2.6 and then thanks to the FlatLaf support team it will work again in the next release. It seems the Proguard problems revolve around that the design that calls in the various UI components by name via the properties file, which makes then unreferenced within the analyzed class usage tree. This causes Proguard to believe they are not used and hence drops those classes from the resulting output jar. As a resolution for this, I have seen other applications explicitly reference their defaults and then use properties as an override mechanism. This will be a future issue if we ever move to native image, such as GraalVM, since it similarly inspects for classes that are explicitly used by the application and by default drops everything else. 

-keep class com.formdev.flatlaf.*, 
	com.formdev.flatlaf.ui.*,
	com.formdev.flatlaf.resources.*, 
	com.formdev.flatlaf.icons.*,
	com.formdev.**.*.properties { *; }

Andy-DG:
[...]

PR #845, opened 2024-05-21
docs: Include a note when using Proguard/Shadow with FlatLaf in README.md

ellet0:
Developers who are using Proguard/Shadow JAR to minimize the jar to reduce its size will get an error when running the final jar which leads to a bug, most developers might not always test the jar minimized by Proguard/Shadow JAR (using a run configuration, for example, script or a task) and caught this bug, even if they do, the error might not be very clear, it's because that Flatlaf use reflection (as [mentioned](https://github.com/JFormDesigner/FlatLaf/issues/648#issuecomment-1441547550)), Swing use it too but the imports already included in Java JRE and will not minimize, this PR add a note for the developers who are using any tool that minimize the jar, and provide one solution which is to exclude FlatLaf from being minimized instead of not minimize the JAR and remove this feature completely, there are other solutions but this can work for more use-cases

If you are interested in getting this into the `README.md` as a note, then feel free to ask me for more changes, like what should The section called and where should it be

Thank you for your efforts in this amazing library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,24 @@ See also
 for instructions on how to redistribute FlatLaf native libraries with your
 application.
 
+Proguard/Shadow JAR
+--------
+
+If you are using [Gradle Shadow JAR Plugin](https://github.com/johnrengelman/shadow) to build the jar for your application and `minimize()` function to minimize the size in the configurations of `shadowJar` task, then you will get an exception `java.lang.Error: no ComponentUI class for:` when running the jar, that's because Swing and Flatlaf both use Reflection, one way to fix this issue without removing the `minimize()` is to exclude Flatlaf from being minimized by using the following:
+
+```groovy
+
+shadowJar {
+    // Your shadow configurations
+    minimize {
+        // Exclude the entire FlatLaf dependency from minimization to fix `no ComponentUI class for: javax.swing.<component>`
+        exclude(dependency("com.formdev:flatlaf:.*"))
+    }
+}
+
+```
+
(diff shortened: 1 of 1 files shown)


ellet0:
Hi @DevCharly, this PR doesn't provide rules/configurations for excluding the classes that use reflections manually. Instead, it provides instructions on how to exclude FlatLaf from minimization by Proguard/Shadow JAR to avoid any runtime errors.

If you update the usage of **reflections** or create more classes that use **reflections**, there will be no need to update this section.

While it seems clear when using **Shadow JAR**, on **Proguard** it requires an additional step that I couldn't find anywhere. It was already in the docs but the details were quite specific and generic, so I didn't know.

I was able to minimize the **open-source** app I'm working on and save space up to more than 60% while still using Flatlaf:

image

Previously, I did a workaround by disabling the FlatLaf themes feature when using Proguard

In case you're interested, feel free to request any changes on this PR, as it will solve the common issues for the most common two minimization tools for the JVM platform.

Thank you once again.

halotroop2288 (approved):

DevCharly:
Have added some basic information regarding obfuscation, shrinking and far JARs to main readme:
https://github.com/JFormDesigner/FlatLaf?tab=readme-ov-file#download

Didn't use your PR because it contains too much information.
The main readme should contain only really necessary information.
Docs on how to use ProGuard/Shadow with Gradle belong to the tools...

Anyway, thanks for your efforts 👍

Issue #860, opened 2024-06-26
Outline of focus component painted with failure


poce1don:
Hi Karl,

See this error that is happening with the outlines in the JScrollPane.

See that in the video the problem does not happen when the scale is at 100% or 150%, but at 125% (my working scale), or 175%, the failure appears.

https://github.com/JFormDesigner/FlatLaf/assets/32749102/e99c1f97-f1b8-446a-af59-58941f33bec9

It doesn't matter the resolution used on the screen, but the scale.

The video screen is my secondary monitor, with 2560x1440 resolution (100% scale). My primary screen is 3840x2160 (125% scale). On both screens, at the mentioned scales, the failure occurs.

In some dimensions where JScrollPane is started, the failure does not occur. In the video, the failures occurred in these specific components, but in JTextField and any other component that displays a border when receiving focus.

Thanks.

DevCharly:
Strange 😕 
Works fine here.

Do you see the same issue in the FlatLaf Demo?

poce1don:
In the demo, this doesn't happen.

I was thinking it was a flaw with GridLayoutManger (IntelliJ), however, with MigLayout the problem was the same.

The tests went up to the 175% scale, and the interesting thing is that it only happens when it is at 125% or 175%, at 100% and 150% they work perfectly, apparently it has something to do with the 25% increases and not with 50%.

I also thought it could be a bug in the JDK and I tested Jetbrains, Bellsoft (my current one), Adoptium and they all had the same problem, with the difference that in one the failure was not visible as soon as the JFrame was shown, but rather when the JScrollPane received the focus.

With your more than 25 years of experience with Swing, do you suspect what could be causing this?

JDK: Bellsoft Liberica JDK 17.0.10
Windows 10 x64 22H2

```java
import com.formdev.flatlaf.FlatDarkLaf;
import com.formdev.flatlaf.FlatLaf;
import net.miginfocom.swing.MigLayout;

import javax.swing.JFrame;
import javax.swing.JList;
import javax.swing.JPanel;
import javax.swing.JScrollPane;
import javax.swing.SwingUtilities;
import javax.swing.WindowConstants;
import java.awt.BorderLayout;
import java.util.Map;

public class LayoutMiglayout extends JFrame {

    public LayoutMiglayout() {
        super("Layout MigLayout");

        var scr1 = new JScrollPane(new JList<String>());
        var pnl1 = new JPanel(new BorderLayout());
[...]

Issue #582, opened 2022-08-23
Tool bar buttons with tab style leave traces on High DPI


Chrriis:
Hi,

While making sure our application works properly on Windows with 125% scale factor, I found a new bug related to tool bar buttons with tab style.

Here is a screenshot:
![ToolBarButtonsHighDpi](https://user-images.githubusercontent.com/1174202/186256610-2249d292-e70a-4b03-97c7-1c1c62dd8513.png)

Button2 is selected and Button1 is hovered. Button5 and Button3 left some traces from previous selections, which is not correct.
Note that when the tooltip disappears, some repainting occurs that might clear some of the traces. I got the gray trace after showing the tooltip and changing selections.

Here is the test case:
[ToolBarButtonTestOnHighDpi.zip](https://github.com/JFormDesigner/FlatLaf/files/9410389/ToolBarButtonTestOnHighDpi.zip)
The button at the bottom allows to increase the top gap, because this problem only happens for certain "y" locations.

It took me hours to isolate that one in a standalone test case! Do not hesitate if you need more information 😉

DevCharly:
Many thanks for the test case 👍 

Looks like a rounding issue in Swing, caused by painting only a small region when e.g. painting button hover.

There are two cases:

1. painting whole window: the full height of the button is filled
2. re-painting only the button: not the full height of the button is filled

![image](https://user-images.githubusercontent.com/5604048/186274482-aa0ac16b-aaa8-4fdb-addf-04380ce5a2a3.png)

In both cases, the same simple fill operation is used: `g.fillRect( 0, 0, c.getWidth(), c.getHeight() );`

Don't yet know what makes the difference... 😕

Chrriis:
Possibly related JDK bugs:
https://bugs.openjdk.org/browse/JDK-8253530
https://bugs.openjdk.org/browse/JDK-8272707

Chrriis:
And the bug I opened: https://bugs.openjdk.org/browse/JDK-8292976

DevCharly:
see PR #864

PR #864, opened 2024-07-05
HiDPI: fix incomplete component repainting at 125% or 175% scaling on Windows

DevCharly:
There is a problem in Swing, when using scale factors that end on .25 or .75
(e.g. 1.25, 1.75, 2.25, etc) and repainting single components.
Sometimes, under special conditions, the right and/or bottom 1px edge of the component is not repainted.

Following screenshot shows examples where the right and/or bottom 1px edge of a component was not repainted:

grafik

This PR introduces new repaint methods that fix/workaround the issue:

~~~java
HiDPIUtils.repaint( Component c );
HiDPIUtils.repaint( Component c, Rectangle r );
HiDPIUtils.repaint( Component c, int x, int y, int width, int height );
~~~

fixes issues #860 and #582

## Repaint Manager

There is also a [repaint manager](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/javax/swing/RepaintManager.html), that also fixes/workarounds the issue,
but does not need any change to existing code (no need to use `HiDPIUtils.repaint()`).
This can be useful for custom or 3rd party components.
Invoke following on application startup to use it:

~~~java
HiDPIUtils.installHiDPIRepaintManager();
~~~

Be careful if already using a custom repaint manager,
or if using a library that may use a custom repaint manager (e.g. SwingX for translucent JXPanel).

## In-depth analysis of the Swing issue

When repainting a component using `Component.repaint()`,
the component is first painted to an in-memory image,
and then that image is copied to the screen.
See [javax.swing.RepaintManager.PaintManager.paintDoubleBufferedFPScales()](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/javax/swing/RepaintManager.java#L1676-L1744).

There are two clipping rectangles involved when copying the image to the screen:
[sun.java2d.SunGraphics2D.devClip](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/sun/java2d/SunGraphics2D.java#L243) and [sun.java2d.SunGraphics2D.usrClip](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/sun/java2d/SunGraphics2D.java#L242).

`devClip` is the device clipping in physical pixels.
It gets the bounds of the painting component, which is either the passed component,
or if it is non-opaque, then the first opaque ancestor of the passed component.
It is calculated in [sun.java2d.SunGraphics2D.constrain()](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/sun/java2d/SunGraphics2D.java#L362-L400) while
getting a graphics context via `JComponent.getGraphics()`.

`usrClip` is the user clipping, which is set via `Graphics` clipping methods.
This is done in [javax.swing.RepaintManager.PaintManager.paintDoubleBufferedFPScales()](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/javax/swing/RepaintManager.java#L1723).

The intersection of `devClip` and `usrClip`
(computed in [sun.java2d.SunGraphics2D.validateCompClip()](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/sun/java2d/SunGraphics2D.java#L1907))
is used to copy the image to the screen.

Unfortunately different scaling/rounding strategies are used to calculate
the two clipping rectangles, which is the reason of the issue.

`devClip` (see [sun.java2d.SunGraphics2D.constrain()](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/sun/java2d/SunGraphics2D.java#L373-L376)):

~~~java
int devX = (int) (x * scale);
int devWidth = Math.round( width * scale )
~~~

`usrClip` (see [javax.swing.RepaintManager.PaintManager.paintDoubleBufferedFPScales()](https://github.com/openjdk/jdk/blob/b83766e59063a41ea8801ac9e7c15dce67727c62/src/java.desktop/share/classes/javax/swing/RepaintManager.java#L1702-L1707)):
~~~java
int usrX = (int) Math.ceil( (x * scale) - 0.5 );
int usrWidth = ((int) Math.ceil( ((x + width) * scale) - 0.5 )) - usrX;
~~~

X/Y coordinates are always rounded down for `devClip,` but rounded up for `usrClip.`
Width/height calculation is also different.

### Conclusion

`devClip` is sometimes 1px too small.

### Workaround

Do repaint on an (larger) ancestor.
In this case, `devClip` is larger because it gets the bounds of the ancestor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ FlatLaf Change Log
 - Theme Editor: Fixed occasional empty window on startup on macOS.
 - FlatLaf window decorations: Fixed black line sometimes painted on top of
   (native) window border on Windows 11. (issue #852)
+- HiDPI: Fixed incomplete component paintings at 125% or 175% scaling on Windows
+  where sometimes a 1px wide area at the right or bottom component edge is not
+  repainted. E.g. ScrollPane focus indicator border. (issues #860 and #582)
 
 #### Incompatibilities
 
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatButtonUI.java
@@ -29,6 +29,7 @@ import java.awt.Graphics;
 import java.awt.Graphics2D;
 import java.awt.Insets;
 import java.awt.Rectangle;
+import java.awt.event.FocusEvent;
 import java.awt.geom.RoundRectangle2D;
 import java.beans.PropertyChangeEvent;
 import java.util.Map;
@@ -61,6 +62,7 @@ import com.formdev.flatlaf.icons.FlatHelpButtonIcon;
 import com.formdev.flatlaf.ui.FlatStylingSupport.Styleable;
 import com.formdev.flatlaf.ui.FlatStylingSupport.StyleableUI;
 import com.formdev.flatlaf.ui.FlatStylingSupport.UnknownStyleException;
+import com.formdev.flatlaf.util.HiDPIUtils;
 import com.formdev.flatlaf.util.LoggingFacade;
 import com.formdev.flatlaf.util.UIScale;
 
@@ -312,11 +314,11 @@ public class FlatButtonUI
 
 			case BUTTON_TYPE:
 				b.revalidate();
-				b.repaint();
+				HiDPIUtils.repaint( b );
 				break;
 
 			case OUTLINE:
-				b.repaint();
+				HiDPIUtils.repaint( b );
 				break;
 
 			case STYLE:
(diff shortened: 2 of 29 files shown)


DevCharly:
Update: added "Repaint Manager" section (see above)

DevCharly:
Update: there is a bug in `HiDPIUtils.installHiDPIRepaintManager()` in 3.5.
Upgrading to 3.5.1 is recommended.

PR #874, opened 2024-07-28
Separate check/radio icon from menu item icon

ShadelessFox:
This is what FlatLaf does, as of now:
sMRyPWdpYJ

It doesn't play well with non-contrast icons and ones that occupy the entire region they're painted on.

This is what I managed to achieve:
vNh6HiOiA5

Also, note that now the entire check/radio icon is painted instead of just the knob part. I quite like how it looks.

This behavior is disabled by default and can be enabled by setting property `MenuItem.separateIconFromCheck` to `true`.


ShadelessFox:
@DevCharly please let me know what you think and if I should change anything :^)

PR #903, opened 2024-10-22
Support for alternate row color in JTree

dar-dev:
Allow settings a custom alternate row color in JTree.  See: https://github.com/JFormDesigner/FlatLaf/issues/900

Cheers,

Dar

--

PR #939, opened 2024-12-13
Support for alternate row color in JList

dar-dev:
Allow setting custom alternate row color in JList as we do for JTable or JTree (see also: https://github.com/JFormDesigner/FlatLaf/pull/903).

Cheers,

Dar

--

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatListUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatListUI.java
@@ -325,7 +325,7 @@ public class FlatListUI
 		}
 		
 		// combobox does not support alternate row color
-		if ( !"ComboBox.list".equals( list.getName() ) ) {
+		if( !"ComboBox.list".equals( list.getName() ) ) {
 			if( alternateRowColor != null && row % 2 != 0 ) {
 				g.setColor( alternateRowColor );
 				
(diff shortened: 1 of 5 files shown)


flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatListUI.java
```diff
@@ -318,6 +323,18 @@ protected void paintCell( Graphics g, int row, Rectangle rowBounds, ListCellRend
 			cx = rowBounds.x;
 			cw = rowBounds.width;
 		}
+		
+		// filelist/combobox does not support alternate row color
+		if( !isFileList && !(dataModel instanceof DefaultComboBoxModel) ) {
```
DevCharly, line 328: Better use `"ComboBox.list".equals( list.getName() )` to check whether list is used in combobox. Not all comboboxes use `DefaultComboBoxModel`. E.g. "Look in" combobox in `JFileChooser`.

Why exclude filelist? Seems to work ok.

DevCharly (changes requested):
Great. Thank you 👍

flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatListUI.java
dar-dev: Just fixed the check and reworked the filelist. 

Thanks

--

PR #971, opened 2025-02-18
Linux: rounded iconify/maximize/close buttons if using FlatLaf window decorations

DevCharly:
This PR changes appearance of iconify/maximize/close buttons **on Linux** from Windows 10/11 style to GNOME style.

New:

image

Old:

image

There are also more possibilities to customize iconify/maximize/close buttons.


iajn:
Would it be possible to opt-in to use the GNOME style buttons on Windows as well? To me they look nicer than the Windows defaults, especially when using the mac look and feel variants.

If not, I suppose copying some of the Linux-specific parts from the default properties files to custom ones would work?

DevCharly:
> If not, I suppose copying some of the Linux-specific parts from the default properties files to custom ones would work?

Yes, that's the way to go 😉

VISTALL:
Hello. Maybe make support prefixes for display env? Like:

```ini
[linux@kde]some.property=1
[linux@gnome]some.property=3
[linux]some.property=0
```

IanKrL:
I am trying to undo these linux-specific changes on my project. What is the simplest way to "unset" the [linux] properties, or set them to the default non-linux-specific values (but on a linux system)? That is, I want to do something like

[linux]TitlePane.closeHoverBackground = $[default]TitlePane.closeHoverBackground

To override the values set in flatlaf-core/src/main/resources/com/formdev/flatlaf/FlatLaf.properties

PR #988, opened 2025-03-22
System File Chooser

DevCharly:
Class `SystemFileChooser` (in package `com.formdev.flatlaf.util`) allows using **operating system file dialogs** in Java Swing applications.

Details: https://www.formdev.com/flatlaf/system-file-chooser/

### Windows

image

### macOS

image

### Linux

image

### Light or Dark ?

On **Windows** and **Linux**, the system file dialogs are shown in **system appearance**. If application uses FlatLaf Dark, but system uses Light appearance, then file dialogs are also shown in Light appearance. Have not found a way to control light/dark appearance on those platforms.

On **macOS**, system file dialogs are shown in same appearance as FlatLaf.  If application uses FlatLaf Dark, but system uses Light appearance, then file dialogs are shown in Dark appearance.

### Background story

Before starting this task, the basic idea was to implement some (small) native
code for Windows and use AWT `FileDialog` on macOS and Linux. Thought that could
be done in 2 or 3 days. I was so wrong...

Turns out that AWT `FileDialog` does not provide necessary features (e.g.
"Select Folder"), window modality does not work, macOS file dialog does not
offer a "file filter" combobox, etc.

So implemented (larger) native code for Windows, macOS and Linux. This was
difficult because I never used GTK before and my macOS Cocoa skills were also limited.

In the end it took **about 4 weeks** to implement, test and document this feature.\
If I had known that, I probably wouldn't have started it...

So if you use FlatLaf in a commercial application, please consider sponsoring
FlatLaf development: https://www.formdev.com/flatlaf/sponsor/

Thank you.


DevCharly:
A snapshot for this PR is now available.
Use `PR-988-SNAPSHOT` as version.
https://github.com/JFormDesigner/FlatLaf#snapshots

Prebuilt binaries (Artifacts) are available here (requires GitHub login):
https://github.com/JFormDesigner/FlatLaf/actions/runs/14009048883

KlemenDEV:
Nice work!

VISTALL:
Hello @DevCharly do you tested it at KDE? How it works ?

ctipper:
I attach crash log, failure is pretty vanilla, just causes segmentation fault when exiting.
[hs_err_pid64597.log](https://github.com/user-attachments/files/20643843/hs_err_pid64597.log)
[hs_err_pid64674.log](https://github.com/user-attachments/files/20643844/hs_err_pid64674.log)

DevCharly:
@ctipper thanks for reporting.

Unfortunately I can not reproduce it 😕 

> ... just causes segmentation fault when exiting

What exactly are you doing?
Open file dialog and then exit app while file dialog is still open?
Or close file dialog and then quit the app?

Does it happen for open, save or select folder file dialogs? Or for all?

Can you reproduce it with the FlatLaf Demo?
(use "File > Open (System)" to open system file dialog)

ctipper:
Exiting the file dialog triggers seg fault. Logs show two different code paths. Not sure I can help much, I thought the logs would be enough. They are both file open dialogs.

DevCharly:
> Exiting the file dialog triggers seg fault.

Does it happen when you press the "Cancel" button? or the "Open" button?

> I thought the logs would be enough

Well, the log contains a lot of Java related information, but not a stack of native calls.
ATM I have no idea where it crashes...

ctipper:
It's the Open button. I'm sorry I've dismantled my test bed. I was getting the system crash reporter but didn't think to save a copy unless you happen to know where in my file system they would be located?

ctipper:
[java-2025-06-08-110041 copy.txt](https://github.com/user-attachments/files/20645563/java-2025-06-08-110041.copy.txt)
[java-2025-06-08-110343 copy.txt](https://github.com/user-attachments/files/20645564/java-2025-06-08-110343.copy.txt)
File format is IPS, I changed file extension manually to upload (unsupported file type)

DevCharly:
> It's the Open button.

And it does not crash if you press the "Cancel" button? Is this correct?

Many thanks for the additional logs. I'll have a look 👍

DevCharly:
Have updated macOS to 15.5 and can now reproduce the issue. Working on a fix...

DevCharly:
The crash on macOS 15.x is now fixed in latest `PR-988-SNAPSHOT`:
https://github.com/JFormDesigner/FlatLaf#snapshots

ctipper:
There don't appear to be any macOS artifacts \@ com.formdev:flatlaf:dylib:macos-arm64:PR-988-SNAPSHOT

ctipper:
Otherwise I can confirm success on arm64 macos (using artifacts rather than snapshots)

ctipper:
I implemented the system file chooser 5 different times, once as a open dialog and 4 times as an save as dialog. One save dialog launched in a minimised state, which needs a trick to maximise. I've only seen this in Preview.app and I don't understand how it's triggered, I didn't do anything special.

ctipper:
Reflecting on the minimised dialog and Preview.app I think I had used this option in my app to save as PDF 

```
fc.addChoosableFileFilter(
    new SystemFileChooser.FileNameExtensionFilter( "PDF Files", "pdf" ) );
```

Could that be the trigger?

Edit: It only happens on first launch, I'm thinking it's just a quirk

DevCharly:
> One save dialog launched in a minimised state, ...

Not sure what you mean by "minimized state" 😕 
Can you please post a screenshot?

> ...which needs a trick to maximise. 

what trick?

ctipper:
image

My app, with an export to PDF dialog. This is minimised. Preview.app presents a minimal version of this when you move a PDF to another folder. This only appears one time, it seems to remember state.

ctipper:
Also wondering what it would take to centre dialog on app window, I only adapted my App quickly and may have missed something.

ctipper:
I have confirmed working in Windows 11. I may be nitpicking but dialogs are wider than my app window and have to be resized.

ctipper:
I've tested in Xfce4 now, all working. I'm also able to do a side by side comparison with JFileChooser and in my opinion the platform needs to incorporate this code, JFileChooser is years old and not fit for purpose. How hard would it be to upstream this PR, I mean it looks really bureaucratic to me but I'm not a JDK dev?

Scoty:
One thing that is missing is passing a `FileSystemView` to the constructor or via a setter method (the`JFileChooser` supports it in a constructor and it can be used to sort the directories and the newest files to be on the top for example).

KlemenDEV:
Not sure if a cross-platform solution for FileSystemView would be possible

DevCharly:
> One thing that is missing is passing a `FileSystemView` to the constructor or via a setter method (the`JFileChooser` supports it in a constructor and it can be used to sort the directories and the newest files to be on the top for example).

@Scoty @KlemenDEV this is not possible because native file dialogs do not support this. They show always the file system.

KlemenDEV:
This is amazing, can't wait to see this in a release! :D

DevCharly:
Just realized that there are some unanswered questions...

> My app, with an export to PDF dialog. This is minimised. Preview.app presents a minimal version of this when you move a PDF to another folder. This only appears one time, it seems to remember state.

@ctipper unfortunately there is no way to control this behavior.
There is a property to get the "expanded" state, but no setter:
https://developer.apple.com/documentation/appkit/nssavepanel/isexpanded

> Also wondering what it would take to centre dialog on app window, I only adapted my App quickly and may have missed something.

@ctipper There is no way to control size or location of native file dialogs. The operating systems decide where to show it and its size.

> I have confirmed working in Windows 11. I may be nitpicking but dialogs are wider than my app window and have to be resized.

@ctipper There is no Windows API to control file dialog size.

DevCharly:
Thanks for testing and the feedback 👍 😃

ctipper:
> PDF dialog. This is minimised

@DevCharly This presented no difficulty as state is preserved, it just appears initially so only mild confusion. Weirdly it seemed to know that it was set for PDF so many users will not discover this

> dialogs are wider than my app window

@DevCharly Let's hope it maintains state, I am not sitting at my test machine

This is all good work, though the Windows feature is the main motivation for me. JFileChooser is not a very useful dialog on Windows. Looking forward to the release.

KlemenDEV:
Once again, thanks for this feature. One step forward to being JavaFX free in our software :D

I have tried to implement the solution for our project (https://github.com/MCreator/MCreator/compare/master...KlemenDEV:MCreator:system_file_chooser) and have noticed one problem.

With e.g. JavaFX that also uses native dailogs, one can defina a file filter such as "_file_suffix.png", so it only shows files that end with suffix + extension, but your file filter prevents dots in extensions and auto-applies dots, preventing this use case. Since JavaFX can do it, I think this should not be limitation of system chooser options. Would it be possible to implement something like this for this library too?

Other than that, seems to work very well, nicely done! :)

KlemenDEV:
Wondering if there is any eta or roadmap for the stable release with this feature? Do you need any help with more testing?

In our testing, outside the feature request above I have no other comments and it seems to work as advertised :)

DevCharly:
Changes in commit 1c6e8774cf1b78654ed6eb0f07179c3bf62149bf:

- fixed missing filter in combobox if only `setFileFilter(myFilter)` is used, but not `addChoosableFileFilter(myFilter)`
- fallback Swing file chooser did ignore `isAcceptAllFileFilterUsed()`
- check for supported filter types in `setFileFilter()`

There are still a few issues open...

DevCharly:
> With e.g. JavaFX that also uses native dailogs, one can defina a file filter such as "_file_suffix.png", so it only shows files that end with suffix + extension...
> Since JavaFX can do it, I think this should not be limitation of system chooser options.

@KlemenDEV 
The problem is macOS. It supports only simple file extensions (without dots in extension).

Also JavaFX on macOS does not support it.
Follow code shows open dialog and allows selection of all .txt files:

```java
FileChooser fc = new FileChooser();
fc.getExtensionFilters().add( new FileChooser.ExtensionFilter( "test", "*_copy.txt" ) );
fc.showOpenDialog( null );
```

When showing save dialog with above filter, then "Save As" field contains "Untitled.txt".

So could offer this feature only for Windows and Linux...
Is this is important for you?

KlemenDEV:
> The problem is macOS. It supports only simple file extensions (without dots in extension).

I see. Honestly, I think there is more value in it working the same on all platforms.

> Also JavaFX on macOS does not support it.

This may actually explain some bug reports we got in the past, was not aware of this :D

> Is this is important for you?

What I did is just made it accept .png and then after file is selected, run additional validation, and if the filename is not right, a pop-up is shown. This is good enough for our use case as it is just one special case where this is needed :)

ArkoSammy12:
Very thankful for this feature. Thank you so much <3

DevCharly:
On Linux GNOME, there is now a good chance that the system file dialog
appears in light/dark system theme depending on FlatLaf theme.
E.g. if system uses light theme, but FlatLaf uses dark theme,
then system file dialog probably is also dark. (was always light before)
But there is no guarantee. This depends on the used/installed system themes.

But this seems not to work on KDE systems...

Numerok:
Is there a way to get chosen filter? 
on my own implementation i used windows HRESULT GetFileTypeIndex(IntByReference ppv) to get the selected filter index including *.*
But with SystemFileChooser when using ch.getFileFilter() i get the first filter and that breaks the logic of choosing the correct report output format for me.
Can it be added or how to do it using this class? i didn't find a way to get it.
Should i open an issue with this request?

DevCharly:
@Numerok this is indeed missing. Please open a new issue. Thanks.

Issue #1006, opened 2025-06-09
can't have JTables without headers


olitank:
Hello,

Unlike the metal look and feel, flatlaf displays headers on a JTable with unnamed columns. 
So you can't have pretty JTables without headers. 
Here's some code that demonstrates the problem:

	package com.otk.jesb;
	
	import javax.swing.JOptionPane;
	import javax.swing.JScrollPane;
	import javax.swing.JTable;
	import javax.swing.SwingUtilities;
	import javax.swing.table.DefaultTableModel;
	
	import com.formdev.flatlaf.FlatLightLaf;
	
	public class TestFlatlafTableHeader {
	
		public static void main(String[] args) {
			SwingUtilities.invokeLater(new Runnable() {
				@Override
				public void run() {
					JOptionPane.showMessageDialog(null,
							new JScrollPane(new JTable(new DefaultTableModel(new Object[] { "" }, 10))));
					FlatLightLaf.setup();
					JOptionPane.showMessageDialog(null,
							new JScrollPane(new JTable(new DefaultTableModel(new Object[] { "" }, 10))));
				}
			});
		}
	
	}
I think I can fix this issue. Is it possible for me to contribute (send a pull request or ...) ?

Regards

DevCharly:
> Unlike the metal look and feel, flatlaf displays headers on a JTable with unnamed columns

Metal also displays header for unnamed columns.
It just has a narrow height:

![Image](https://github.com/user-attachments/assets/250d5de6-7050-4454-a34d-faa597a0850e)

> Is it possible for me to contribute (send a pull request or ...) ?

Sure

olitank:
Hello, 

Can you grant me permissions for the pull request ?
Or tell me what else should I do ?
I could not push my branch (https://github.com/JFormDesigner/FlatLaf.git: git-receive-pack not permitted on 'https://github.com/JFormDesigner/FlatLaf.git/').

Regards

DevCharly:
You need to fork this repo to your github account (click "Fork" button here https://github.com/JFormDesigner/FlatLaf).

Then clone your fork to the local computer,
create a new branch for the change,
make the change and test it,
commit and push it to your fork.

Then you can create a PR on GitHub.

more details: https://docs.github.com/en/pull-requests

DevCharly:
Turns out that there is a property `TableHeader.height = 25` that specifies the preferred height of table header.
See https://www.formdev.com/flatlaf/components/tableheader/#sizes

[...]

PR #1010, opened 2025-06-23
JTable empty header size (too high) fix

olitank:
The issue: https://github.com/JFormDesigner/FlatLaf/issues/1006
I just modified the method **com.formdev.flatlaf.ui.FlatTableHeaderUI.getPreferredSize(JComponent)**.
After debugging, I found that the header height (size.height) was correctly calculated by super.getPreferredSize(c). I thought getPreferredSize(JComponent c) had been overridden solely to adapt the header height to the flatlaf.uiScale property. 
Can you confirm this?

I then deduced that this was a typo and that simply replacing "UIScale.scale(height)" with "UIScale.scale(size.height)" would fix this small issue.

I also added: flatlaf-testing/src/main/java/com/formdev/flatlaf/testing/FlatEmptyTableHeaderTest.java to verify the fix. But I'm not sure if this is really useful.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTableHeaderUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTableHeaderUI.java
@@ -300,7 +300,7 @@ public class FlatTableHeaderUI
 	public Dimension getPreferredSize( JComponent c ) {
 		Dimension size = super.getPreferredSize( c );
 		if( size.height > 0 )
-			size.height = Math.max( size.height, UIScale.scale( height ) );
+			size.height = Math.max( size.height, UIScale.scale( size.height ) );
 		return size;
 	}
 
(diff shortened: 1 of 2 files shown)


flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatTableHeaderUI.java
```diff
@@ -300,7 +300,7 @@ protected void paintBottomSeparator( Graphics g, JComponent c, int x, int w ) {
 	public Dimension getPreferredSize( JComponent c ) {
 		Dimension size = super.getPreferredSize( c );
 		if( size.height > 0 )
-			size.height = Math.max( size.height, UIScale.scale( height ) );
+			size.height = Math.max( size.height, UIScale.scale( size.height ) );
```
DevCharly, line 303: This does not look right...
It is actually the same as `size.height = UIScale.scale( size.height )`, isn't it?
But `size.height` is already scaled.
And preferred height (field `height`) is no longer used...

DevCharly:
Turns out that there is a property `TableHeader.height = 25` that specifies the preferred height of table header.
See https://www.formdev.com/flatlaf/components/tableheader/#sizes

You can set this to zero, then it works as in other L&Fs:

~~~java
UIManager.put( "TableHeader.height", 0 );
~~~

Issue #413, opened 2021-11-03
Is it possible to scale a single component?


dvtomas:
I'd like to have some elements of my app (in my particular case a few major check-boxes and radio-buttons) bigger (that is, bigger font, bigger check box, check mark, etc). Is it possible to achieve that somehow? [This](https://docs.oracle.com/javase/tutorial/uiswing/lookandfeel/size.html) link mentions using `component.putClientProperty("JComponent.sizeVariant", "large")`, but it doesn't seem to do anything with my checkbox.

I need this only for particular instances, that is, *not* for all checkboxes or for all radio buttons.

DevCharly:
There is no general way to scale a single components.
The client property `JComponent.sizeVariant` works only in Aqua and Nimbus L&Fs and is not supported by FlatLaf.

Anyway, for checkboxes and radiobuttons you're in luck. I've implemented zoomable icons last week for the FlatLaf theme editor (code is not yet in git repo):

![image](https://user-images.githubusercontent.com/5604048/140094799-f308464f-5d4c-4b42-8553-cc8ffe267909.png)

~~~java
private static class ZoomCheckBoxIcon
    extends FlatCheckBoxIcon
{
    private final float zoom = 1.3f;

    @Override
    public void paintIcon( Component c, Graphics g, int x, int y ) {
        Graphics2D g2 = (Graphics2D) g.create();
        try {
            g2.translate( x, y );
            g2.scale( zoom, zoom );

            super.paintIcon( c, g2, 0, 0 );
        } finally {
            g2.dispose();
        }
    }

    @Override
    public int getIconWidth() {
        return Math.round( super.getIconWidth() * zoom );
    }

    @Override
    public int getIconHeight() {
        return Math.round( super.getIconHeight() * zoom );
    }
}
~~~

For radiobutton icon, duplicate above class and extend `FlatRadioButtonIcon`.

Usage:

~~~java
checkBox1.setIcon( new ZoomCheckBoxIcon() );
Font font = checkBox1.getFont();
checkBox1.setFont( font.deriveFont( font.getSize() * 1.3f ) );
~~~

![image](https://user-images.githubusercontent.com/5604048/140095550-ffd33376-6c08-4066-ae84-59339064ce44.png)

dvtomas:
Amazing, thank you so much!

dvtomas:
Hmm, there's a problem that such explicitly set icon doesn't respect LaF change. Is there any way around that?

[...]

PR #1060, opened 2025-11-15
Styling of wrong icon in FlatRadioButtonUI

daWoife:
The method applyStyleProperty of class FlatRadioButtonUI currently styles the variable icon, a field of the parent class BasicRadioButtonUI and the default icon which is set with UIManager.getIcon in method installDefaults.  
So the wrong icon is styled if someone subclasses FlatRadioButtonIcon or FlatCheckBoxIcon and sets this as the new icon for a JRadioButton or JCheckBox instance. (An example why someone would do so is shown in issue #413). 
With this change styling takes account of the current icon of a JRadioButton or JCheckBox.
Maybe the methods getStyleableInfos and getStyleableValue should also take account of the current icon, but this is not done in this change.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatRadioButtonUI.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/ui/FlatRadioButtonUI.java
@@ -203,16 +203,12 @@ public class FlatRadioButtonUI
 	protected Object applyStyleProperty( AbstractButton b, String key, Object value ) {
 		// style icon
 		if( key.startsWith( "icon." ) ) {
-			Icon styleIcon = b.getIcon();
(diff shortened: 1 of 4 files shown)


DevCharly:
Thanks for the PR, but it did not work for me because variable `styleIcon` was not used to apply style.
I've fixed it and also updated methods `getStyleableInfos()` and `getStyleableValue()`.
`getIconFocusWidth()`, and therefore `getPreferredSize()` did also use the wrong icon.

daWoife:
Thanks for your changes, 
You are right, the line where applyStyleProperty is called musst use the new variable styleIcon instead of icon in my code. 
I used the Web-Interface to make the PR manually, so I forgot to change this important line.

PR #1070, opened 2025-12-28
Add Cairo font module

hyasm:
Adds Cairo font family as a new FlatLaf fonts module.

- Includes all available Cairo weights
- Uses same structure and API as existing font modules
- Font licensed under SIL Open Font License 1.1


DevCharly:
I'm sorry, but I don't want to add the Cairo font to the FlatLaf.
I've nothing against the font itself.

The idea of having some fonts in FlatLaf was to make it easier to use a few very popular fonts in Swing.
There are thousands of fonts out there and I don't want to add more to FlatLaf at the moment.

If you think that Cairo font is useful for many Swing applications/developers, you could create your own independent GitHub project.

If you need it in your application, simply add the files of this pull request to your application project.

PR #1074, opened 2025-12-31
Recalculate geometry after styling in FlatSliderUI

daWoife:
The parent class BasicSliderUI caches it's calculated geometry for the layout of the track, thumb, ticks, etc.
So it is necessary to recalculate the geometry in FlatSliderUI if the slider is styled to have a correct preferred size.
It's a simple solution without checking what exactly is changed, the track width, the thumb size or only a color.
But at least it checks the internal oldStyleValues map, so if it is empty before and after the styling no recalculation is done.


DevCharly:
Many thanks 👍

PR #1080, opened 2026-01-13
Fix NPE in FlatUIDefaultsInspector when color is null

poce1don:
The UI Defaults Inspector renders color values by painting the table cell background with the corresponding color, which makes color identification immediate and visual.

However, when listing `FlatLineBorder` values, although the border properties `(insets`, `color hex value` and `line thickness`) are displayed, the value cell background was not painted using the border color. This made it difficult to visually identify the actual color used by the border.

This change extends the existing behavior so that, when a `FlatLineBorder` is displayed, the value cell background is painted using the border’s line color, providing the same visual feedback already available for plain color values.

According to the FlatLaf documentation, borders defined using the
`top,left,bottom,right[,lineColor[,lineThickness[,arc]]] `format allow `lineColor`, `lineThickness`, and `arc` to be optional.

This means definitions such as `5,5,5,5,,,15`, specified in theme `.properties` files, are valid and supported, resulting in a rounded border without a line color, which works correctly when applied to Swing components.

When such a border is encountered by the UI Defaults Inspector, the absence of a `lineColor` results in a null value, which previously caused a `NullPointerException` during rendering.

This change ensures that the inspector handles this documented and valid configuration gracefully, avoiding runtime exceptions and improving robustness and usability when inspecting theme-defined borders.

The implementation is minimal and does not affect existing rendering behavior for other value types.

### DIFF
--- a/flatlaf-extras/src/main/java/com/formdev/flatlaf/extras/FlatUIDefaultsInspector.java
+++ b/flatlaf-extras/src/main/java/com/formdev/flatlaf/extras/FlatUIDefaultsInspector.java
@@ -777,6 +777,9 @@ public class FlatUIDefaultsInspector
 
 		@SuppressWarnings( "FormatString" ) // Error Prone
 		private static String color2hex( Color color ) {
+
+			if ( color == null ) return "#none";
+
 			int rgb = color.getRGB();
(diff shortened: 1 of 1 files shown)


DevCharly:
Many thanks 👍

PR #1087, opened 2026-01-24
Fix exception when SVG icon name has no file extension

poce1don:
When using dark themes, FlatSVGIcon attempts to insert `_dark` before the file
extension. If the icon name does not contain a dot, this results in a
`StringIndexOutOfBoundsException.`

This situation can easily occur if the SVG file path is provided without the
`.svg` extension by mistake, since the API does not enforce or validate the
file name format.

This change adds a simple guard to avoid invalid substring operations when
no file extension is present, preventing runtime exceptions while preserving
existing behavior.


DevCharly:
Merged in latest `3.7.1-SNAPSHOT`: https://github.com/JFormDesigner/FlatLaf#snapshots

PR #1110, opened 2026-04-01
TextField: fixed preferred size when leading/trailing component or icon are present

DJ-Raven:
This PR fixes an issue with TextField margins when using leading/trailing components or icons.

When no layout constraint is set and the JTextField column is 0, the layout uses the component’s preferred size. In this case, an extra horizontal margin becomes visible.

From reviewing the existing implementation, the left and right insets are reduced when leading/trailing components are present, but this is not reflected in the preferred size calculation.

This PR resolves the issue by updating `getPreferredSize(JComponent c)` to adjusted insets.

Before Fix

no-fix

After Fix

fix


DevCharly:
Many thanks 👍

PR #1142, opened 2026-08-28
macOS: Add support for the Cmd+Backspace and Cmd+Delete text shortcuts in textboxes

Omegabird113:
On macOS, `cmd` + `backspace` is a standard shortcut for deleting to the line beginning and `cmd` + `delete` is an equivalent to the line end.

Swing does not natively support these shortcuts, so this PR adds custom swing action definitions inside of `FlatInputMaps.initTextComponentInputMaps()` and maps those to `meta` + `BACK_SPACE`/`DELETE`.

### DIFF
--- a/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatInputMaps.java
+++ b/flatlaf-core/src/main/java/com/formdev/flatlaf/FlatInputMaps.java
@@ -16,16 +16,16 @@
 
 package com.formdev.flatlaf;
 
-import javax.swing.InputMap;
-import javax.swing.JTextField;
-import javax.swing.KeyStroke;
-import javax.swing.LookAndFeel;
-import javax.swing.UIDefaults;
+import javax.swing.*;
 import javax.swing.UIDefaults.LazyValue;
-import javax.swing.UIManager;
 import javax.swing.plaf.InputMapUIResource;
+import javax.swing.text.BadLocationException;
+import javax.swing.text.JTextComponent;
+import javax.swing.text.Utilities;
 import com.formdev.flatlaf.util.SystemInfo;
 import static javax.swing.text.DefaultEditorKit.*;
+import java.awt.*;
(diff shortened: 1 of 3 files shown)


Omegabird113:
Conflicts resolved. I'll keep trying to resolve conflicts (mostly in `CHANGELOG.md`) if they come up when I notice that they happened. When I first tried to merge, I botched it, so I re-did it and force-pushed there.

flatlaf-core/src/main/java/com/formdev/flatlaf/FlatInputMaps.java
```diff
@@ -172,6 +177,54 @@ private static void initTextComponentInputMaps( UIDefaults defaults ) {
 			"control shift O", "toggle-componentOrientation", // DefaultEditorKit.toggleComponentOrientation
 		};
 
+		AbstractAction deleteToBeginLineAction = new AbstractAction()
```
DevCharly: The implementation of the two actions do not fit into this class.
 
- better move it as nested (static) classes to the end of class `FlatTextFieldUI`.
- extend `TextAction` (as done in other text actions)
- use `TextAction` method `getTextComponent(e)` to get text component (as done in other text actions)

DevCharly (changes requested):

flatlaf-core/src/main/java/com/formdev/flatlaf/FlatInputMaps.java
Omegabird113: > The implementation of the two actions do not fit into this class.
> 
> * better move it as nested (static) classes to the end of class `FlatTextFieldUI`.
> * extend `TextAction` (as done in other text actions)
> * use `TextAction` method `getTextComponent(e)` to get text component (as done in other text actions)

Done

DevCharly:
thank you very much 👍

